PR #4, opened 2018-10-16
Implement UB1 practical exercises

### LINKED ISSUE #1: implement first exercise
https://pp.ipd.kit.edu/lehre/WS201819/compprakt/intern/blatt1.pdf

problame:
fixes #1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,5 +26,4 @@ script:
   # let build fail on Clippy warnings
   # and also check tests and none-default crate features
   - cargo clippy --all-targets --all-features -- -D warnings
-  - cargo build
-  - cargo check
+  - cargo test
--- a/build
+++ b/build
@@ -12,12 +12,53 @@ done
 cd "`dirname "${SCRIPT_PATH}"`" > /dev/null
 SCRIPT_PATH="`pwd`";
 popd  > /dev/null
-echo "change working directory to '$SCRIPT_PATH'"
+echo "[INFO] change working directory to '$SCRIPT_PATH'"
 cd "$SCRIPT_PATH"
 
-set -x
+# default config for speedcenter
+BUILD_FLAGS="--release"
+DO_CLEAN=1
+DO_CLIPPY=0
+DO_BUILD=1
+DO_TEST=0
+DO_CHECK=0
+while [ "$#" -ne 0 ]; do
+case "$1" in
+    "--speedcenter") ;;
+    "--noclean") DO_CLEAN=0 ;;
+    "--debug")  BUILD_FLAGS="" ;;
+    "--release")  BUILD_FLAGS="--release" ;;
+    "--ci") # shortcut to emulate CI build
+        BUILD_FLAGS=""
+        DO_CLEAN=0
+        DO_CLIPPY=1
+        DO_BUILD=0
+        DO_TEST=1
+        DO_CHECK=0
+        ;;
+    *)
+        echo "[ERROR] invalid argument: $1" >&2
+        exit 1
+        ;;
+esac
+shift
+done
 
-cargo clippy --all-targets --all-features -- -D warnings
-# NOTE: for --release builds also change ./run to actually use it!
-cargo build
-cargo check
+run_if() {
+    cond="$1"
+    shift
+    if [ "$cond" = 1 ]; then
+        echo "$@"
+        eval "$@"
+    else
+        echo "[SKIP] $@"
+    fi
+}
+
+run_if "$DO_CLEAN"  cargo clean
+# let build fail on Clippy warnings
+run_if "$DO_CLIPPY" cargo clippy $BUILD_FLAGS --all-targets --all-features -- -D warnings
+run_if "$DO_BUILD"  cargo build  $BUILD_FLAGS
+run_if "$DO_TEST"   cargo test   $BUILD_FLAGS
+# and also check tests and none-default crate features
+run_if "$DO_CHECK"  cargo check  $BUILD_FLAGS
--- a/build
+++ b/build
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -ex
+cargo clippy --all-targets --all-features -- -D warnings
+# NOTE: for --release builds also change ./run to actually use it!
+cargo build
+cargo check
--- a/build
+++ b/build
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 set -e
 
 # change to this script's directory
--- a/build
+++ b/build
@@ -1,5 +1,22 @@
 #!/bin/sh
-set -ex
+set -e
+
+# change to this script's directory
+# credits to https://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself/12197518#12197518
+pushd . > /dev/null
+SCRIPT_PATH="${BASH_SOURCE[0]}";
+while([ -h "${SCRIPT_PATH}" ]); do
+    cd "`dirname "${SCRIPT_PATH}"`"
+    SCRIPT_PATH="$(readlink "`basename "${SCRIPT_PATH}"`")";
+done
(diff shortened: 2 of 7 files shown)


build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cargo clippy --all-targets --all-features -- -D warnings
+# NOTE: for --release builds also change ./run to actually use it!
```
flip1995: We could make the build and run script always build and use release versions and add additional scripts if we want build scripts for ourselves. If we have to modify these scripts every week before a release sounds like a bad idea IMO.

reiner-dolp:
If I read this correctly the run script is not yet independent of the current working directory, right?

image

build
problame: If it's only used for speedcenter builds then yes.
On the other hand, I find it handy to have a local script to check whether CI will complain about a commit I make (before pushing).

problame:
I think it boils down to *do not change cwd because we will pass you relative path names* (?)

build
reiner-dolp: we could just make an additional script `run_debug`?

reiner-dolp:
But won't `exec relative_path_to_compiler_binary` fail?

build
flip1995: Yes I can see this. We could add a `build-debug` script for this. This would allow to also skip some steps that are maybe necessary on a release, but not in development. Or the other way around.

E.g. for the speedcenter we don't need a Clippy check but _I_ want to enforce this here.

joshuabach:
Shouldn't `exec ${0%/*}/target/debug/comprakt` do it? I think we can assume that `$0` contains the full path to the script

joshuabach:
And isn't this a problem in the build script also? E.g. `cargo build` won't work if the cwd is not our project dir

build
```diff
@@ -0,0 +1,23 @@
...
+popd  > /dev/null
+echo "change working directory to '$SCRIPT_PATH'"
+cd "$SCRIPT_PATH"
+
+set -x
+
+cargo clippy --all-targets --all-features -- -D warnings
+# NOTE: for --release builds also change ./run to actually use it!
+cargo build
+cargo check
```
reiner-dolp: @flip1995 `cargo build` already performs a `cargo check`, right? And doesn't `cargo clippy` performs a `cargo build`? so are these two compiler invocations necessary?

flip1995: `cargo build` performs `cargo check`. `cargo clippy` builds all dependencies, but only checks the crate IIRC.

Yes looks like it:
https://github.com/rust-lang-nursery/rust-clippy/blob/6ae89c4f1123d81e56141b02cd3b83436c1ad035/src/driver.rs#L88-L92

flip1995: But theoretically 
```
cargo clippy
cargo test
```
should be enough

joshuabach: I think `check` is in fact covered by `build` (or `clippy`), but `clippy` doesn't seem to invoke `build`:
```
 $ cargo clean
 $ ./build
change working directory to '/home/joshua/repos/by-owner/comprakt/comprakt'
+ cargo clippy --all-targets --all-features -- -D warnings
   Compiling proc-macro2 v0.4.20
   Compiling unicode-xid v0.1.0
    Checking libc v0.2.43
    Checking unicode-width v0.1.5
    Checking vec_map v0.8.1
    Checking ansi_term v0.11.0
    Checking strsim v0.7.0
    Checking bitflags v1.0.4
    Checking textwrap v0.10.0
    Checking atty v0.2.11
    Checking clap v2.32.0
   Compiling quote v0.6.8
   Compiling syn v0.15.11
   Compiling structopt-derive v0.2.12
    Checking structopt v0.2.12
    Checking comprakt v0.0.1 (/home/joshua/repos/by-owner/comprakt/comprakt)
    Finished dev [unoptimized + debuginfo] target(s) in 15.01s
+ cargo build
   Compiling libc v0.2.43
   Compiling unicode-width v0.1.5
   Compiling bitflags v1.0.4
   Compiling vec_map v0.8.1
   Compiling strsim v0.7.0
   Compiling ansi_term v0.11.0
   Compiling textwrap v0.10.0
   Compiling atty v0.2.11
   Compiling clap v2.32.0
   Compiling structopt v0.2.12
   Compiling comprakt v0.0.1 (/home/joshua/repos/by-owner/comprakt/comprakt)
    Finished dev [unoptimized + debuginfo] target(s) in 14.02s
+ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
```

src/main.rs
```diff
@@ -1,3 +1,28 @@
+extern crate structopt;
```
flip1995: We're using the 2018 edition, so `extern crate` isn't necessary anymore.

Could you enable the lint group `#![warn(rust_2018_idioms)]` so we're pushed to use the 2018 edition to the fullest?

build
reiner-dolp: ok, so only the `cargo check` is unnecessary.

run
```diff
@@ -13,4 +13,16 @@ cd "`dirname "${SCRIPT_PATH}"`" > /dev/null
...
 
-exec "$SCRIPT_PATH/target/debug/comprakt" $@
+target="release"
+while [ "$#" -ne 0 ]; do
+case "$1" in
+    "--debug")
+        target="debug"
+        ;;
+    *)
+        echo "invalid argument: $1" >&2
```
joshuabach: Wait but if I now to `run --echo Foo.java`, it prints "invalid argument", right? Thats not good...

.travis.yml
```diff
@@ -26,5 +26,4 @@ script:
   # let build fail on Clippy warnings
   # and also check tests and none-default crate features
   - cargo clippy --all-targets --all-features -- -D warnings
-  - cargo build
-  - cargo check
+  - cargo test
```
problame: We could also call `./build --ci` and be done with it. But then Travis loses the info about the individual build steps...

run
problame: -,- can't seem to get anything done properly today

.travis.yml
flip1995: IDK then we would execute all the useless commands we need for their CI. I would keep travis seperated from this

problame: `./build --ci` refers to emulation of travis, not speedcenter

reiner-dolp: bikeshedding!

flip1995: Yes but finding the script_path in travis is not necessary at all. This would just clutter the travis log. Our travis log will probably grow really big sooner or later, so running as few commands as possible and having the info about the steps is a advantage IMO.

build
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
problame: last bikeshed: we should use bash here, since we rely on bashisms.
However, use `/bin/bash` or `/usr/bin/env bash`?
Since speedcenter is Debian Linux, we can be pretty sure bash will be installed to `/bin/bash`.

joshuabach: I always use "/usr/bin/env bash", since its more portable. But in this case it dosen't matter. We could of course vote on slack ;-)

src/main.rs
```diff
@@ -1,3 +1,29 @@
...
+
+use structopt::StructOpt;
+
+use std::path::PathBuf;
+
+#[derive(StructOpt)]
+#[structopt(name="comprakt")]
+struct Opt {
+    #[structopt(long="echo")]
+    echo: bool,
```
reiner-dolp: sidenote: comments added to the variants will show up in `--help`.
```rs
/// writes the input file to stdout without modification 
#[structopt(long="echo")]
echo: bool,
```

flip1995:
sorry I want to try something :smile: 

bors r-

flip1995:
Well 4/5 approvals. Let's try this:

bors r=flip1995,joshuabach,reiner-dolp

flip1995:
> 4: Implement UB1 practical exercises r=flip1995,joshuabach,reiner-dolp a=problame

Nice it works!

PR #8, opened 2018-10-19
Add tasksheet 1

joshuabach:
The theoretical tasks from the first sheet

### DIFF
--- a/tasksheets/sheet1.md
+++ b/tasksheets/sheet1.md
@@ -1,0 +1,46 @@
+Task 1
+======
+We use Rust. For details on the infrastructure, refer to the README.
+
+
+Task 2
+======
+Done.
+
+
+Task 3
+======
+
+(1) MiniJava Program
+----------------
+```java
+class Foo {
+    public Foo[] bar;
+
+    public static void main(String[] args) {
+        System.out.println(42);
+    }
+}
+```
+
+(2) On the Turing-Completeness of MiniJava
+--------------------------------------
+MiniJava is turing complete, since it provides both conditional expressions (*IfStatement*),
+as well as an iteration construct (*WhileStatement*). Unlimited memory is achieved via the
+32-bit `int` type as a local variable and using recursion (*Method* and *MethodInvocation*)
+to "allocate" more memory. Since there is no mechanism for getting the address of a variable
+(like e.g. `&varname` in C), the size of the Stack is not bounded, and can grow arbitrarily
+(at least in theory).
+
+(3) Subset of Java
+--------------
+Yes, because the grammar yields a sublanguage of that of Java, and for the semantics only
+additional constraints are given to the semantics of Java, so any valid MiniJava program
+is a valid Java program.
+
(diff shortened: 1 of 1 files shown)


tasksheets/sheet1.md
```diff
@@ -0,0 +1,53 @@
...
+
+(3) Subset of Java
+--------------
+Yes, because the ignored characters ("Leerraum"), the Comment-patterns and the keywords
+(<=SE8, see note below) are a subset of those of Java; because the grammar yields a sublanguage
+of that of Java; and for the semantics only additional constraints are given to the semantics
+of Java, so any valid MiniJava program is a valid Java (<=SE8) program.
+
+However, since Java release 9 [1], the underscore `_` is a reserved keyword, which it is not
+in MiniJava, so that MiniJava is no longer a subset of Java.
```
problame: ... subset of Java **9 and newer**

tasksheets/sheet1.md
```diff
@@ -0,0 +1,53 @@
...
+}
+```
+
+(2) On the Turing-Completeness of MiniJava
+--------------------------------------
+MiniJava is turing complete, since it provides both conditional expressions (*IfStatement*),
+as well as an iteration construct (*WhileStatement*). Unlimited memory is achieved via the
+32-bit `int` type as a local variable and using recursion (*Method* and *MethodInvocation*)
+to "allocate" more memory. Since there is no mechanism for getting the address of a variable
+(like e.g. `&varname` in C), the size of the Stack is not bounded, and can grow arbitrarily
```
problame: Took me a few seconds to understand why this implication holds...

joshuabach: Reading it again, my formulation is not even true. The implication only holds in the opposite direction (addressable memory => bounded memory). There could be other "features" of the language that I missed which impose bounds on the memory... Proving turing-completeness is really hard ;-) I'll try to reformulate and be a little more verbose

tasksheets/sheet1.md
```diff
@@ -0,0 +1,54 @@
...
+
+(3) Subset of Java
+--------------
+Yes (of Java <= SE8), because the ignored characters ("Leerraum"), the Comment-patterns and the keywords
+(at least for Java <= SE8, see note below) are a subset of those of Java; because the grammar yields a sublanguage
+of that of Java; and for the semantics only additional constraints are given to the semantics
+of Java, so any valid MiniJava program is a valid Java (<= SE8) program.
+
+However, since Java SE9 [1], the underscore `_` is a reserved keyword, which it is not
+in MiniJava, so that MiniJava not a subset of Java >= SE9.
```
flip1995: ... which is not part of MiniJava, so MiniJava is not a subset of Java >= SE9

problame: @reiner-dolp please add the example and ref to the standard, would consider this discussion resolved then

flip1995:
Please **always** use `bors` to merge PRs.

PR #11, opened 2018-10-24
first draft of string table

reiner-dolp:
we need to add interior mutability for the string table to be usable beyond the lexer stage.

Contains unsafe code, needs a detailed review.


problame:
I would be fine lke it is now, too.
I started integrating StringTable with the Lexer but had some troubles with lifetimes:
anonymous lifetime of  `&mut self` in the **lexer** methods != lifetime of the `&'t mut StringTable` member, and since the `&'t mut StringTable` member of lexer can't be used by value, the lifetime during the call `self.symtab.intern(...)` is shorter than the lifetime required for the result of that call.

I had to rebase this branch onto the lexer branch, my current state is in branch https://github.com/comprakt/comprakt/tree/problame-string-interner-into-lexer
 The words above probably make more sense if you look at the latest commit (and try compile it, get the error)

PR #12, opened 2018-10-24
Lexer

joshuabach:
Implement lexer. This uses a `Rc`-based string table, we might wan't to use @reiner-dolp's instead, see #11.

This implements all the suggestions made directly to commits in the original lexer branch.

ToDo:
- [ ] Diagnostics 
     - [x] Errors
     - [ ] Warnings (e.g. `/*` inside comment)
     - [ ] Pretty printing
- [x] Token display (currently just line-by-line `Debug` and the string table are dumped to stdout)
- [x] Rewrite `lex_operator`. This function is really ugly
- [x] Post processing (`.filter` comments + whitespace, etc)
- [ ] Unit tests
- [x] Apply @flip1995's fuzzer
- [ ] Comments & Inline-doc


joshuabach:
I suggest including error handling / diagnostics in the post-processing, as that is less invase to the existing lexer, and also would make it easier to later on "igonore" errors to see how far we can get (to display more diagnostics

hediet:
Schön soweit :) Ich kommentiere morgen mal weiter.

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    fn lex_whitespace(&mut self) -> Token<'t> {
+        self.lex_while(|c| c.is_whitespace(), |_, _, _| TokenData::Whitespace)
+    }
+
+    #[allow(clippy::cyclomatic_complexity)]
+    fn lex_operator(&mut self) -> Option<Token<'t>> {
+        use self::Operator::*;
+
+        match self.input.peek_multiple(4) {
```
hediet: Warum nicht andersherum testen? Zuerst das erste Zeichen testen und bei Uneindeutigkeit weiter das nächste Zeichen testen anstatt von vorne anzufangen?

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
...
+}
+
+impl TryFrom<&str> for Keyword {
+    type Error = ();
+
+    fn try_from(s: &str) -> Result<Keyword, Self::Error> {
+        use self::Keyword::*;
+
+        match s {
+            "abstract" => Ok(Abstract),
```
hediet: Sicher dass du das nicht generieren willst? :D

src/main.rs
```diff
@@ -62,7 +72,20 @@ fn run_compiler(cmd: &CliCommand) -> Result<(), Error> {
...
+                File::open(&path).context(CliError::OpenInputError { path: path.clone() })?;
+            let mapping = (unsafe { Mmap::map(&file) })
+                .context(CliError::MmapError { path: path.clone() })?;
+            let ascii_file = asciifile::AsciiFile::new(mapping).unwrap();
+
+            let strtab = strtab::StringTable::new();
+            let lexer = lexer::Lexer::new(ascii_file.iter(), &strtab);
+
+            for t in lexer {
+                println!("{:?}", t.data);
```
hediet: Whitespaces sollten doch eigentlich nicht als token ausgegeben werden. Trotzdem find ich es sinnvoll, die im tokenizer zu berücksichtigen.

joshuabach: "Post Processing" steht als ToDo im PR ;-)

src/strtab.rs
```diff
@@ -0,0 +1,41 @@
...
+impl StringTable {
+    pub fn new() -> Self {
+        StringTable {
+            entries: RefCell::new(HashSet::new()),
+        }
+    }
+
+    pub fn intern(&self, value: String) -> Symbol<'_> {
+        let value = Rc::new(value);
+        self.entries.borrow_mut().insert(Rc::clone(&value));
```
reiner-dolp: This will replace the old value on each new intern() call. Effectively interning nothing!!!

src/lexer.rs
joshuabach: Ich habe gedacht, dass das dann so ein hässlich verzweigter Baum wird. Zugegebenermaßen ist das jetzt auch nicht viel schöner, aber zumindest kann man bei jedem Token auf den ersten Blick die zugehörige Zeichenkette erkennen, bei dem anderen müsste man immer den match-baum zurückverfolgen um das herauszufinden

hediet: Darum -> Generator :D

src/asciifile.rs
```diff
@@ -0,0 +1,277 @@
+use failure::Fail;
+use memmap::Mmap;
+
+pub struct AsciiFile {
+    mapping: Mmap,
+}
+
+#[derive(Debug, Fail)]
+pub enum EncodingError {
```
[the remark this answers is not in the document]
joshuabach: Maybe we get more/different encoding errors in the future ... (we probably won't)

src/asciifile.rs
```diff
@@ -0,0 +1,277 @@
...
+        self.peek_multiple(1).chars().next()
+    }
+
+    pub fn eof_reached(&mut self) -> bool {
+        self.peek().is_some()
+    }
+
+    pub fn eof_position(&mut self) -> Position {
+        if self.next().is_some() {
+            panic!("Must only be called at EOF")
```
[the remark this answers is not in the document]
joshuabach: Well, this is only called in situations where the caller is sure that we are at EOF, so i would just `.unwrap()` and write `// Unwrap is safe because we are at EOF` anyway

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
+use crate::{
+    asciifile::{Position, PositionedChar, PositionedChars},
+    strtab::*,
+};
+
+use std::{convert::TryFrom, result::Result};
+
+macro_rules! match_op {
+    ($input: expr, $len: expr, $right: expr) => {{
+        // Unwraps are safe, because this is only called after token of length $len,
```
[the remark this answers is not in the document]
joshuabach: I guess that would be the "Ansatz 2" suggested on the slides. I don't think it would be that hard to implement this inside the `PositionedChars` iterator (i already have a buffer for the peeked values), just turn this around a bit. 

This would be more problematic in the lexer itself, I think: If e.g. the `.peek()` in line 253 (`lex_token`) would already consume the first digit of the integer, `lex_integer_literal` would miss it's first digit. There are probably equally many arguments in the code for both sides.

But turning it around would get rid of the (safe) unwraps... Hm.

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
...
+            let PositionedChar(pos, _) = $input.next().unwrap();
+            end = pos;
+        }
+
+        Some(Token::new(begin, end, TokenData::Operator($right)))
+    }};
+}
+
+#[derive(Debug)]
+pub enum TokenData<'t> {
```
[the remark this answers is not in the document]
joshuabach: But doesn't `Kind` indicate that it's just the kind/type, and not the data?

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
...
+    Keyword(Keyword),
+    Operator(Operator),
+    Identifier(Symbol<'t>),
+    IntegerLiteral(Symbol<'t>),
+    Comment(String),
+    UnclosedComment(String),
+    UnexpectedCharacter(char),
```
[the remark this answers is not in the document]
joshuabach: This was easier in the beginning, which is why I went for it. It would probably be nicer to make the lexer an iterator over `Result<Token, LexicalError>`, where `LexicalError` contains variants for unclosed comments & unexpected characters.

src/strtab.rs
joshuabach: Nope. `HashSet::insert` calls `HashMap::insert` which, according to https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert 
> If the map did have this key present, the value is updated, and the old value is returned. The key is not updated, though; this matters for types that can be == without being identical.

So, since the keys are the Strings, the old value won't be replaced. The values are just `()`.

joshuabach: Your interner still saves more allocation, of course (or has the potential too, with a more thrifty lexer)

reiner-dolp: Oh.... I misread the docs. In this case this is actually pretty ingenious :)

src/asciifile.rs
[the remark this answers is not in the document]
reiner-dolp: It's a public function and can be called from anywhere.

src/asciifile.rs
```diff
@@ -0,0 +1,277 @@
...
+
+#[derive(Debug, PartialEq)]
+pub struct PositionedChar(pub Position, pub char);
+
+pub struct PositionedChars<I>
+where
+    I: Iterator<Item = char>,
+{
+    curpos: Position,
+    af: I, // form ascii file
```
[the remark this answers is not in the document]
flip1995: Yeah if I read af I think of "as f***" and not at all about a ascii file. I would recommend naming it `ascii_file`

src/lexer.rs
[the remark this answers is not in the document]
joshuabach: Also returning the cursor position in `peek()` would make the lexer code a bit more obfuscated, avoiding this was the originial reason I added a custom `peek` function. At the very least, almost every peek would have to be trailed with `.map(|v| v.2)`.

src/lexer.rs
```diff
@@ -0,0 +1,432 @@
...
+    fn lex_whitespace(&mut self) -> Token<'t> {
+        self.lex_while(|c| c.is_whitespace(), |_, _, _| TokenData::Whitespace)
+    }
+
+    #[allow(clippy::cyclomatic_complexity)]
+    fn lex_operator(&mut self) -> Option<Token<'t>> {
+        use self::Operator::*;
+
+        match self.input.peek_multiple(4) {
+            ">>>=" => match_op!(self.input, 4, TripleRightChevronEqual),
```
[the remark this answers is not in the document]
flip1995: I can think of 2 ways, but both are kind of ugly:
```rust
match {
    x if x == ">>>=" => match_op!(self.input, x.len(), _)
    ...
}
```

A macro:
```rust
macro_rules! wrapper_macro {
    ($input:expr, $(($token_string:ident, $token:expr))+, $len:expr, $else:expr) ...
} 
```

src/lexer.rs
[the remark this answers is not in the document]
flip1995: But the enum variants are just the `Kind` of the `Token`. The data is in the fields of the variants.

Another suggestion: Just name this `Token` and the current `Token`-struct `TokenAndSpan` (this idea is not at all stolen from [rustc](https://doc.rust-lang.org/nightly/nightly-rustc/syntax/parse/lexer/struct.TokenAndSpan.html) :smile: )

src/lexer.rs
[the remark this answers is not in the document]
reiner-dolp: Would we be able to use the macro as follows?
```
match peek() {
macro!(">>>=", self.input, TripleRightChevronEqual)
macro!("<<<=", self.input, TripleLeftChevronEqual)
/*...*/
}
```

That would be how chrome/webkit lexes HTML.

src/lexer.rs
flip1995: I like this version. It's not pretty, but it's easy to understand and hopefully Just Works™

src/lexer.rs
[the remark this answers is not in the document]
flip1995: Yes such a macro would be possible
```rust
macro_rules! macro {
    ($input:expr, $( ($token_string:ident, $token:expr) ),+) => {
        $(
            stringify!($token_string) => match_op!($input, stringify!($token_string).len(), $token),
        )+
    }
}
```

and then
```rust
match peek_multiple(4) {
    macro!(self.input, (>>>=, TripleRightChevronEqual))
    _ => match peek_multiple(3) {
        macro!(self.input, 
              (<<=, DoubleLeftChevronEqual),
              (>>=, DoubleRightChevronEqual),
              (>>>, TripleRightChevron),
        )
    }
}

```

src/lexer.rs
joshuabach: It will, however, get even uglier because implementing `Display` for `Keyword` and `Opetaror` will require to basically reimplement `Keyword::TryFrom<&str>` and `lex_operator` in reverese. Not very DRY...

src/main.rs
```diff
@@ -122,3 +114,104 @@ fn print_error(writer: &mut dyn io::Write, err: &Error) -> Result<(), Error> {
...
+#[cfg(test)]
+mod lexertest_tests {
+
+    macro_rules! lexer_test_with_tokens {
+        ( $toks:expr ) => {{
+            let v: Vec<TokenData<'_>> = { $toks };
+            let mut o = Vec::new();
+            let res = run_lexer_test(v.into_iter(), &mut o);
+            assert!(res.is_ok());
+            String::from_utf8(o).expect("output mut be utf8")
```
flip1995: `mut`->`must`

src/lexer.rs
```diff
@@ -34,6 +33,24 @@ pub enum TokenData<'t> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use self::TokenData::*;
+
+        match self {
+            Keyword(keyword) => write!(f, "{}", keyword),
+            Operator(operator) => write!(f, "{}", operator),
+            Identifier(symbol) => write!(f, "identifier {}", symbol),
+            IntegerLiteral(symbol) => write!(f, "integer literal {}", symbol),
+            Comment(body) => write!(f, "/*{}*/", body),
+            UnclosedComment(body) => write!(f, "//{}", body),
```
flip1995: `UnclosedComment` is not `//`, but `/*`, without `*/`. Do we want to print this as `//`?

reiner-dolp: If I understand correctly, UnclosedComments are not allowed, therefore this should be removed by @joshuabach patch that will also remove UnexpectedCharacter.

joshuabach: `UnclosedComment` (now in `ErrorKind`) does not mean a `// comment`, but reaching end of file while parsing a `/* comment` ... So yes, it's an error now

src/main.rs
```diff
@@ -53,16 +69,28 @@ fn main() {
...
+            let ascii_file = asciifile::AsciiFile::new(mapping)
+                .context(CliError::Ascii { path: path.clone() })?;
+
+            let strtab = strtab::StringTable::new();
+            let lexer = lexer::Lexer::new(ascii_file.iter(), &strtab);
+
+            let mut stdout = io::stdout();
+            // TOOD without vector? Maybe itertools::process_results? Or for-loop and
+            // early-return?
+            let tokens: Result<Vec<_>, _> = lexer.collect();
```
problame: should not collect

src/main.rs
```diff
@@ -82,14 +110,113 @@ fn print_error(writer: &mut dyn io::Write, err: &Error) -> Result<(), Error> {
     let mut causes = err.iter_chain();
 
     if let Some(err_msg) = causes.next() {
-        writeln!(writer, "Error: {}", err_msg)?;
+        writeln!(writer, "error: {}", err_msg)?;
     } else {
-        writeln!(writer, "Unknown Error")?;
+        writeln!(writer, "Unknown error")?;
```
problame: can causes ever have length 0?

reiner-dolp:
No. But writing it like this instead of unwrapping does not hurt.

src/main.rs
joshuabach: > No. But writing it like this instead of unwrapping does not hurt.

I think it does hurt, since it suggests that it might be the case. Leaving out the else case ore putting `unreachable!()` is more meaningful.

flip1995: +1 for `unreachable!()`

flip1995:
```
error: failed to mmap MiniJava file "/home/pkrones/Documents/comprakt/mjtest/tests/lexer/0770c3a293e1439c93431d6e423ef09a.mj"
caused by: memory map must have a non-zero length
```
hab noch einen Fehler gefunden!

problame:
Doesn't minijava require a main method? Anyways, that's no concern to asciifile, so it's a bug

flip1995:
Looking at the grammar an empty file is syntactically correct:

image

flip1995:
Empty files with `Mmap` are a real problem. I figured out this ugly solution:
```rust
let mapping = (unsafe { Mmap::map(&file) })
    .or_else(|err| {
        if let Ok(metadata) = file.metadata() {
            if metadata.len() == 0 && io::ErrorKind::InvalidInput == err.kind() {
                let empty_map = MmapOptions::new().len(1).map_anon();
                if let Ok(mut map) = empty_map {
                    map.copy_from_slice(b" ");
                    return map.make_read_only();
                }
            }
        }
        Err(err)
    })
    .context(CliError::Mmap { path: path.clone() })?;
```

I hope someone has a better idea

joshuabach:
I think we should merge this. There are some things still missing, but this is a working lexer that technically satisfies the conditions by the tasksheet, so anything else is better suited to a seperate merge request.

joshuabach:
> Empty files with `Mmap` are a real problem. I figured out this ugly solution:
> [...]
> I hope someone has a better idea

Instead of an `Mmap`, we pass `&[u8]` to `AsciiFile::new` (which is trivial, since `Mmap` is essentially only a `Deref<Target=&[u8]>`). Than you can simply return an empty `&[0; u8]` from your `or_else` block, or sth similar.

src/lexer.rs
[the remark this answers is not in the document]
problame: orthogonal, but I'm afraid my upcoming commit of the following diff will close this comment:
```
diff --git a/src/lexer.rs b/src/lexer.rs
index 75c7164..0d591c9 100644
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -17,14 +17,10 @@ macro_rules! match_op {
     ($input:expr, $len:expr, $right:expr) => {{
         // Unwraps are safe, because this is only called after token of length $len,
         // is already matched and thus contained in $input
-        let PositionedChar(begin, _) = $input.next().unwrap();
-        let mut end = begin;
-        #[allow(clippy::reverse_range_loop)] // Macro might be called with with $len=1
-        for _ in 1..$len {
-            let PositionedChar(pos, _) = $input.next().unwrap();
-            end = pos;
-        }
-
+        assert!($len >= 1);
+        let mut it = $input.by_ref().take($len);
+        let PositionedChar(begin, _) = it.next().unwrap();
+        let end = it.last().map(|pc| pc.0).unwrap_or(begin);
         Some(Ok(Token::new(begin, end, TokenKind::Operator($right))))
     }};
 }
```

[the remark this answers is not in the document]
problame: ok, github was clever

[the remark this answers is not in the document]
problame: As far as I see it, this conversation is the only one left. May the stakeholders decide...

src/asciifile.rs
```diff
@@ -65,7 +65,7 @@ use std::fmt::{self, Display};
 
 impl Display for Position {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(fmt, "{}:{}", self.row, self.col)
+        write!(fmt, "{}:{}", self.row + 1, self.col + 1)
```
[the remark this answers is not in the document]
hediet: IDEs do this as well: Lines/Characters start with 1.

src/main.rs
```diff
@@ -53,16 +69,44 @@ fn main() {
...
+            let mmres = unsafe { Mmap::map(&file) };
+            const EMPTY: [u8; 0] = [0; 0];
+            let bytes: &[u8] = match mmres {
+                Ok(ref m) => m,
+                Err(e) if e.kind() == io::ErrorKind::InvalidInput => {
+                    // Linux returns EINVAL on file size 0, but let's be sure
+                    let file_size = file
+                        .metadata()
+                        .map(|m| m.len())
+                        .context("could not get file metadata while interpreting mmap erorr")?;
```
flip1995:
```suggestion
                        .context("could not get file metadata while interpreting mmap error")?;
```

src/main.rs
```diff
@@ -79,17 +123,108 @@ fn exit_with_error(err: &Error) -> ! {
...
+        writeln!(out, "{}", td)?;
     }
 
     Ok(())
 }
+
+#[cfg(test)]
+mod lexertest_tests {
+
+    macro_rules! lexer_test_with_tokens {
```
flip1995: As @hediet  pointed out on slack, this could also be a function, instead of a macro. This macro does nothing, a function couldn't do. (A function would even be more concise)

flip1995:
After a final look over the code, I think this is good to go after a few minor changes:

- Use `debug_assert!` instead of `assert!`, for release performance
- Use less asserts, but handle the cases covered by the `assert!`s in another way, if possible
- The unresolved comments above
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r227985562
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r228714329
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r228714493
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r228714535
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r228728849
  - [ ] https://github.com/comprakt/comprakt/pull/12#discussion_r228729229

joshuabach:
>     * Use `debug_assert!` instead of `assert!`, for release performance

The Rust documentation suggest using `debug_assert!` only for *really* expensive test, and not by default: https://doc.rust-lang.org/nightly/std/macro.debug_assert.html#uses

>     * Use less asserts, but handle the cases covered by the `assert!`s in another way, if possible

Most of the `asserts!` in the lexer verify that a `lex_<stuff>` function is only called when there really is a beginning of `<stuff>` at point; or to verify in some other way, that the current state of the token stream is as expected. I don't think we can't get rid of these without extensive redesign (including moving away even further from the DFA Idea).

I was even planning to introduce more `asserts!`, as a check before each "safe" `.unwrap()`.

joshuabach:
> > really expensive test, and not by default
> 
> Rust doc: `"that are too expensive to be present in a release build"`
> 
> Yes I read that to, but as you pointed out most of the asserts check if the lexer is in the right state, which involves `peek` or `multiple_peek`, which most of the time results in peeking twice for the same thing. IMO doing stuff twice, without any benefit is "too expensive" for the release build.

I actually don't think the peeks are that expensive, since in case of a second peek for the same length, a slice into the peek buffer is simply returned. So a second peek is just a integer-comparision and a str-subslicing. But you can use `debug_assert!`, n a release binary, all the `asserts!` should succed anyway, if we did enough testing ;-)

flip1995:
In the rust compiler even integer comparisons are done with `debug_assert`s instead of asserts:
https://github.com/rust-lang/rust/search?q=debug_assert&unscoped_q=debug_assert

> in a release binary, all the asserts! should succeed anyway

Hopefully yes.

src/lexer.rs
```diff
@@ -512,7 +512,7 @@ where
     }
 
     fn lex_comment(&mut self) -> TokenResult {
-        assert_eq!(self.input.peek_multiple(2), "/*");
+        debug_assert!(self.input.peek_multiple(2) == "/*");
```
joshuabach: Use https://doc.rust-lang.org/std/macro.debug_assert_eq.html, for better failure reporting

flip1995: Oh this exists? I searched for it, but couldn't find it. oops :smile:

flip1995:
bors r=flip1995,problame,joshuabach

PR #16, opened 2018-10-27
Experiment: Lexer with string slices and absolutely no string allocation/copy

joshuabach:
The lexer in this branch works by looking at string slices instead of iterating over characters, thus saving some allocation. For further testing the strtab here is a stub that just returns the very same `&str`, because I wanted to get rid of any copy. Thus, all strings in the token stream (except comments) are actually references into the `Mmap`.

My approach is based on a "Frame API", where a frame is a portion of the input stream that we are currently *looking at*; and a set of functions to (conditionally) extend and finalize a given frame.

The performance impact is nonexistent for the *holyidentifier* test: 
```
 $ perf stat target/release/comprakt  --lextest ../../IPDSnelting/mjtest-tests/lexer/holyidentifier.mj > /dev/null

 Performance counter stats for 'target/release/comprakt --lextest ../../IPDSnelting/mjtest-tests/lexer/holyidentifier.mj':

        289.778556      task-clock (msec)         #    0.896 CPUs utilized
                80      context-switches          #    0.276 K/sec
                 3      cpu-migrations            #    0.010 K/sec
             5,946      page-faults               #    0.021 M/sec
       924,356,555      cycles                    #    3.190 GHz                      (83.22%)
       236,365,751      stalled-cycles-frontend   #   25.57% frontend cycles idle     (83.49%)
        86,733,737      stalled-cycles-backend    #    9.38% backend cycles idle      (66.42%)
     2,105,200,828      instructions              #    2.28  insn per cycle
                                                  #    0.11  stalled cycles per insn  (83.03%)
       424,070,446      branches                  # 1463.429 M/sec                    (83.88%)
         2,188,197      branch-misses             #    0.52% of all branches          (83.00%)

       0.323576051 seconds time elapsed

       0.257247000 seconds user
       0.032405000 seconds sys

```

From a SW-engineering perspective, this is a two-sided sword. On the one hand I got rid of most of the `unwraps`, (and can also get rid of the rest, I think). On the other hand, the Frame API requires a lot of str- and index-fiddling internally (though I think I get rid of some of that), which is more error-prone than a simple char iterator.

Haven't updated the test, do not compile.

### DIFF
--- a/src/asciifile.rs
+++ b/src/asciifile.rs
@@ -1,6 +1,10 @@
 use failure::Fail;
 use memmap::Mmap;
 
+use std::ops::Deref;
+
+use std::fmt::{self, Display};
+
 pub struct AsciiFile {
     mapping: Mmap,
 }
@@ -17,7 +21,7 @@ pub enum EncodingError {
 
 const ENCODING_ERROR_MAX_CONTEXT_LEN: usize = 180;
 
-impl<'a> AsciiFile {
+impl AsciiFile {
     // cost: O(fileLen) since we need to check if all chars are ASCII
     pub fn new(mapping: Mmap) -> Result<AsciiFile, EncodingError> {
         if let Some(position) = mapping.iter().position(|c| !c.is_ascii()) {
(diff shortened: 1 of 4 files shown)


problame:
> The performance impact is nonexistent for the holyidentifier test:

Could you giv

joshuabach:
> Perf-wise, this patch is a pretty significant regression on my machine.

That's weird. Well, I don't think I have the time this weekend to investigate this, so if no one else want's to get on it, we should leave it as it is

joshuabach:
Closed in favor of #46

PR #18, opened 2018-10-29
Warnings and Errors

reiner-dolp:
- Adds a new Diagnostic object that prints nice error messages and warnings
- Do not terminate the program on the first error
- Bundles state that is shared between lexer, parser and the semantic analysis in a `Context` object

image
image


src/asciifile.rs
```diff
@@ -16,22 +18,62 @@ pub enum EncodingError {
...
+    /// If the cursor/byte offset is on a newline character, the newline
+    /// character will belong to the previous line, meaning the return
+    /// value will not be the given by the byte
+    /// offset, but the byte offset of the previous new line.
+    fn get_line_start_idx(mapping: &'m [u8], byte_offset: usize) -> usize {
+        let region_start = byte_offset
+            .checked_sub(ENCODING_ERROR_MAX_CONTEXT_LEN)
+            .unwrap_or(0);
+
+        let region = &mapping[region_start..byte_offset];
```
flip1995: This could panic if `byte_offset > mapping.len()`. The same could happen in the function `get_line_end_idx` above.

reiner-dolp: Fixed in commit: [reiner-diagnostics f733f22] add debug_assert to byte_offset methods

reiner-dolp:
With error/warning context truncation set to 5 chars instead of 80:

image

src/asciifile.rs
```diff
@@ -33,17 +34,31 @@ impl<'m> AsciiFile<'m> {
...
-            .min(byte_offset + ENCODING_ERROR_MAX_CONTEXT_LEN);
+    fn get_line_end_idx(
+        mapping: &'m [u8],
+        byte_offset: usize,
+        max_context_length: usize,
+    ) -> (LineContext, usize) {
+        debug_assert!(byte_offset <= mapping.len());
+
+        let region_max_end = byte_offset + max_context_length;
+        let (truncation, region_end) = if mapping.len() > region_max_end {
```
flip1995: `let region_end = mapping.len().min(region_max_end);`

src/asciifile.rs
```diff
@@ -33,17 +34,31 @@ impl<'m> AsciiFile<'m> {
...
+        let newline_position = region
             .iter()
             .position(|&chr| chr as char == '\n')
-            .map(|pos| pos + byte_offset)
-            .unwrap_or(region_end)
+            .map(|pos| pos + byte_offset);
+
+        match newline_position {
+            Some(position) => (LineContext::NotTruncated, position),
+            None => (truncation, region_end),
```
flip1995:
```rust
None if region_max_end > mapping.len() => (LineContext::NotTruncated, region_end),
None => (LineContext::Truncated, region_end),
```

src/diagnostics.rs
```diff
@@ -1,29 +1,19 @@
-//! This is a list of semantic and lexical errors and warnings the compiler
-//! emits.
-//!
-//! Error numbers could be generated automatically, _however_ error numbers
-//! should be consistent for all versions of the compiler, even if errors or
-//! warnings are retired.
+//! The diagnostics object controls the output of warnings and errors generated
+//! by the compiler during the lexing, parsing and semantic analysis phases.
+//! The diagnostics object is tracks the number of warnings and errors
```
flip1995: ~~is~~

flip1995:
r=me with prettified function and typo fix

src/asciifile.rs
reiner-dolp: won't fix.

reiner-dolp:
Emitting newlines in an error renders weirdly
```rs
Error: unexpected character '
'
```

reiner-dolp:
Fixed in commit b1e26ca, but the line printing still echos bells and other weirdness:

```
echo -e "a\a bb" > bell.mj
```
image

reiner-dolp:
Bugs to fix:
- [ ] formatter fails for lines containing tabs because tab width != 1
- [ ] formatter echos control sequences unprocessed.

flip1995:
Can we use [`escape_default`](https://doc.rust-lang.org/std/primitive.char.html#method.escape_default) for the escaping of characters?

reiner-dolp:
Yes. Seems to be better than casting to u8 and calling ascii::escape_default

reiner-dolp:
I am reverting, using unicode escape sequences for ASCII files looks weird.

reiner-dolp:
image

src/main.rs
```diff
@@ -240,3 +240,48 @@ mod lexertest_tests {
...
+        let expected_path = format!("{}/{}", INTEGRATION_TEST_DIR, filename);
+
+        let mut expected_file =
+            File::open(&expected_path).expect(&format!("file does not exist: {}", expected_path));
+
+        let mut expected_stderr = String::new();
+        expected_file.read_to_string(&mut expected_stderr).unwrap();
+
+        // do not use color
+        // TODO: Environment() empty does not work for some reason
```
reiner-dolp: `Environment::empty()`

flip1995:
IMO We should merge the diagnostic stuff now and do the integration testing in a different PR.

reiner-dolp:
Removed through force push, moved to branch reiner-continous-integration

src/diagnostics.rs
```diff
@@ -0,0 +1,392 @@
...
+        output.set_color(None);
+        writeln!(output.writer(), "{}", self.kind.as_fail());
+    }
+
+    fn write_colored_with_code<'ctx>(
+        &self,
+        writer: &mut dyn WriteColor,
+        span: &Span,
+        file: &AsciiFile<'ctx>,
+    ) {
```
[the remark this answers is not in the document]
reiner-dolp: will be part of a separate pull request. I don't want to refactor the ASCII drawing code without integration tests verifying stderr did not change.

src/asciifile.rs
```diff
@@ -58,6 +134,81 @@ use std::ops::Deref;
...
+
+impl Position {
+    pub fn get_line<'m>(
+        &self,
+        file: &AsciiFile<'m>,
+        max_context_length_before: usize,
+        max_context_length_after: usize,
+    ) -> (LineContext, &'m str, LineContext) {
+        let (truncated_before, start) = AsciiFile::get_line_start_idx(
+            file.mapping,
```
problame: A `Position` obtained from a different `AsciiFile` might cause the `debug_assert` in `get_line_start_idx` to panic.

src/asciifile.rs
```diff
@@ -58,6 +134,81 @@ use std::ops::Deref;
...
+            self.byte_offset,
+            max_context_length_before,
+        );
+        let (truncated_after, end) =
+            AsciiFile::get_line_end_idx(file.mapping, self.byte_offset, max_context_length_after);
```
problame: A `Position` obtained from a different `AsciiFile` might cause the `debug_assert` in `get_line_end_idx to panic`.

src/diagnostics.rs
```diff
@@ -0,0 +1,461 @@
+//! The diagnostics object controls the output of warnings and errors generated
+//! by the compiler during the lexing, parsing and semantic analysis phases.
+//! It also tracks the number of warnings and errors generated for flow control.
+//!
+//! This implementation is NOT thread-safe.
```
problame: Isn't this redundant given that `RefCell` is `!Sync` and thus any `#[derive(Send,Sync)]` will fail anyways?

src/diagnostics.rs
```diff
@@ -0,0 +1,461 @@
...
+/// is not color aware will print everything in the
+/// color last used.
+impl<'a> Drop for ColorOutput<'a> {
+    fn drop(&mut self) {
+        // ignore coloring failures using ok()
+        self.writer.reset().ok();
+    }
+}
+
+const MAX_CONTEXT_LENGTH: usize = 80;
```
problame: given we use `120` (?) length, and even `180` might be common, I would advocate for `180` here.

src/main.rs
```diff
@@ -130,37 +159,23 @@ fn print_error(writer: &mut dyn io::Write, err: &Error) -> Result<(), Error> {
     Ok(())
 }
 
-fn run_lexer_test<L, O>(lexer: L, out: &mut O) -> Result<(), Error>
```
problame: This function was split up into `run_compiler` (looping) and `write_token` (filtering whitespace and comments, and `writeln`).

In consequence, `lexer_test_with_tokens`had to be fixed by duplicating the iteration code there (`tokens.into_iter()`).
This makes it a white-box test, whereas it was black-box test before.

I assume this was done because `--lexertest` now has two outputs (`stdout` and the `Context.write`).
The new `run_lexer_test` should have 2 writers as parameters.

src/context.rs
```diff
@@ -0,0 +1,35 @@
...
+impl<'ctx> Context<'ctx> {
+    pub fn new(file: AsciiFile<'ctx>, writer: Box<dyn WriteColor>) -> Self {
+        Self {
+            file,
+            strtab: StringTable::new(),
+            diagnostics: Diagnostics::new(writer),
+        }
+    }
+
+    pub fn default(file: AsciiFile<'ctx>) -> Self {
```
problame: The exercises require a particular output behavior. Although it is limited to `stdout`, and this default is for `stderr`, I would rather have `write` injected for any instantiation of `Context`.
Although it's unlikely that there will be a requirement on `stderr` output format, I would rather have an overview of where `stderr` is written to right from `main.rs:run_compiler()`

src/main.rs
```diff
@@ -97,22 +104,44 @@ fn run_compiler(cmd: &CliCommand) -> Result<(), Error> {
                 Err(e) => Err(e.context(CliError::Mmap { path: path.clone() }))?,
             };
 
+            let mut stdout = io::stdout();
+
             let ascii_file = asciifile::AsciiFile::new(&bytes)
                 .context(CliError::Ascii { path: path.clone() })?;
+            let context = Context::default(ascii_file);
```
problame: We want to explicitly control the write of `Context` from here, please remove the default to `stderr`.

problame:
The main complaints I have:
* `Context` should not have a default constructor since program output is a hard requirement and should be tightly controlled right from `run_compiler`
* `Position` should have a reference to the AsciiFile it originated from, and `get_line` should be implemented using that reference.
  * In practice, this bloats the `Position` struct and is unnecessary for our use case (always a single file), but it improves correctness

src/diagnostics.rs
reiner-dolp: Just to clarify: This is a soft limit (since its defined on bytes instead of number of columns required to render the bytes) and per side relativ to the error position.

Since its only there to deal with generated code with a few hundred chars per line, pumping this to 120 might be a good idea, but could result in (120*2+1)*4 columns being displayed in the worst case (for a long line of tab characters and the middle tab character being the error position)

src/main.rs
reiner-dolp: IMO the `run_lexer_test` is weird. @flip1995 had the same opinion. It only abstracts away a simple for loop.

PR #20, opened 2018-10-30
WIP: Parser (Tasksheet 3)

reiner-dolp:
# TODO

## Required before merge
- [x] Error Handling
- [x] Operator Precedence Table
- [x] Handling of `else`

## Goodies
- [x] CLI Interface
- [ ] (Try to) get rid of explicit type ascriptions
- [x] Refactor `Parser::{next, peek, peek_multiple}`. These are not very DRY
- [ ] `Error::expected` is a `String`. Can do better?
- [x] Testing, Testing, Testing!


src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+
+        self.read_eof();
+    }
+
+    fn parse_class_declaration(&self) {
+        self.read_keyword(Keyword.Class);
+        self.read_identifier();
+
+        self.read_op(Operator.LeftBracket);
+        loop {
```
flip1995: Same as above.

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+
+            self.parse_class_member();
+        }
+        self.read_op(Operator.RightBracket);
+    }
+
+    fn parse_class_member(&self) {
+        self.read_keyword(Keyword.Public);
+
+        self.try_read_keyword(Keyword.Static);
```
flip1995: This function is not defined. I would recommend to make a branch here to a function `parse_main_method` for easier AST generation later.

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.parse_type();
+        self.read_identifier();
+
+        if self.try_peek(Operator.LeftParen) {
+            self.read(Operator.RightParen);
```
flip1995: That can't be right. Shouldn't this be `eat(LeftParen)`?

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+
+            // method or main method
```
flip1995: This won't work. We need many more checks for the main method. -> _Is the param `String[] IDENT` the only param_?

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+            loop {
```
flip1995: Again same as above

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+                // parameters
+                if self.try_peek(Operator.RightParen) { break; }
+                self.parse_parameter();
+            }
+
+            self.read(Operator.RightParen);
+
+            if (self.try_read_keyword(Keyword.Throws)) {
```
flip1995: This is completely unconsistent. A `eat(Throws)` needs to follow here!

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+        while self.try_peek_op(Operator.LeftBrace) { // [
+            self.read_op(Operator.LeftBrace);
+            self.read_op(Operator.RightBrace);
+            // Array Type
+        }
+    }
+
+    fn parse_basic_type(&self) {
+        if self.try_peek_keyword(Keyword.Int) {
+            self.try_peek_keyword(Keyword.Int);
```
flip1995: `eat_keyword()`

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+        }
+        else {
+            // named type
+            self.read_identifier();
+        }
+    }
+
+    fn parse_block(&self) {
+        self.read_op(Operator.LeftBracket);
+        loop {
```
flip1995: Again, same as above.

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+            // empty block
+            self.read_op(Operator.Semicolon);
+        }
+        else if self.try_peek_keyword(Keyword.If) {
+            self.read_keyword(Keyword.If);
+            self.read_op(Operator.LeftParen);
+            self.parse_expression();
+            self.read_op(Operator.RightParen);
+
+            self.parse_statement();
```
flip1995: What about else? Have you thought about the dangling else problem?

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn parse_expression(&self) {
+        parse_binary_expression(-1)
+    }
+
+    fn parse_binary_expression(&self, cur_priority: u32) {
+        parse_expression();
+
+        // tries to read some binary operators
+        loop {
+            if let Op(op, op_priority, associativity) = self.try_peek_binary_op() {
```
flip1995: What is Op?

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+                if (op_priority > priority) break;
```
flip1995: Not rust syntax

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+                self.read_op(op);
+                let expr2 = self.parse_binary_expression(op_priority);
+            }
+            else {
+                break;
+            }
+        }
+    }
+
+    fn parse_unary_expr(&self) {
```
flip1995: What about unary expressions? This function is never called. 

I can't tell if this function is correct...

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn parse_binary_expression(&self, cur_priority: u32) {
+        parse_expression();
+
+        // tries to read some binary operators
+        loop {
+            if let Op(op, op_priority, associativity) = self.try_peek_binary_op() {
+                if (op_priority > priority) break;
+                
+                self.read_op(op);
+                let expr2 = self.parse_binary_expression(op_priority);
```
flip1995: This needs way more thought for the AST generation. We need to parse left **and** right hand side of a binary expr before building the bin op.

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+            self.read_op(Operator.LeftParen);
+            self.parse_expression();
+            self.read(Operator.RightParen);
+        }
+        else if self.try_peek_keyword(Keyword.New) {
+            self.read_keyword(Keyword.New);
+
+            self.parse_basic_type();
+
+            if true {
```
flip1995: I'm not yet sure how to implement this nicely, but we cannot do this.

I would at least introduce another function `parse_new_expr`, since this is/will be non trivial logic.

src/parser.rs
```diff
@@ -0,0 +1,287 @@
...
+            }
+        }
+        else {
+            self.fail();
+        }
+    }
+
+    fn parse_parenthesized_argument_list() {
+        self.read_op(Operator.LeftParen);
+        loop {
```
flip1995: That won't work, because of commata. I think the same error has `parse_parameters()`

joshuabach:
A few Notes: 
- At night, I get weird ideas about how to model things, resulting in this generic-type-massacre. Sorry. ;-)
   On the other hand, this way everything works with only one `unwrap`, and looks really nice... Apart from the explicit type ascriptions on many method calls. Fuck You Rust-Type-Inference-Algorithm.

- The Last "big" commit today was 0c64e7d, and its WIP, because It was written quickly and looks ugly. Will refactor.

- There is no real error handling, and I just realized I havn't filled out the operator precedence table. These shortcomings are documented by failing test (that should work once the table and the `unimplemented!` bits are filled)

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+        } else if self
+            .omnomnoptional::<Exactly, _>(Operator::LeftParen)?
+            .is_some()
+        {
+            // parenthesized expression
+            self.parse_expression()?;
+            self.omnomnom::<Exactly, _>(Operator::RightParen)?;
+
+            Ok(())
+        } else if self.omnomnoptional::<Exactly, _>(Keyword::New)?.is_some() {
```
flip1995: we should put this code block in it's own parse function

src/main.rs
```diff
@@ -5,6 +5,10 @@
 #![feature(const_str_as_bytes)]
 #![feature(box_syntax)]
 #![feature(repeat_generic_slice)]
+#![feature(never_type)]
+
+#[macro_use]
+extern crate derive_more;
```
flip1995: This extern crate is probably needed for the `#[display(..)]` attrs?

grammer.md
```diff
@@ -0,0 +1,73 @@
+# The Grammer as is currently implemented
+```
+Program                             ClassDeclaration*
+ClassDeclaration                    class IDENT { ClassMember* }
+ClassMember                         Field
+                                    | Method
+Field                               public Type IDENT ;
+// TODO: MainMethod missing
```
flip1995: We should differ between main/static methods and normal methods already in the parser/AST. If we find a `static` keyword we just continue in a `parse_main_method`-function. In the AST this will then be expressed by a `ClassMemberKind` enum.

src/main.rs
```diff
@@ -162,6 +112,110 @@ fn print_error(writer: &mut dyn io::Write, err: &Error) -> Result<(), Error> {
...
+
+    let mut parser = Parser::new(unforgiving_lexer);
+
+    // TODO: implement diagnostics on parser
+    //if context.diagnostics.errored() {
+    //context.diagnostics.write_statistics();
+    //exit(1);
+    //}
+
+    match parser.parse() {
```
flip1995: We need to filter the `Whitespace` and `Comment` Tokens before calling this

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+            // TODO Not tail recursive
+            self.parse_binary_expression(prec)?;
+        }
+
+        Ok(())
+    }
+
+    fn parse_unary_expression(&mut self) -> ParserResult<'f> {
+        // try read prefixOp
+        if self.omnomnoptional::<UnaryOp, _>(UnaryOp)?.is_some() {
```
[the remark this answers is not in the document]
flip1995: > The lexer produces MinusMinus for -- but we need this to be 2 tokens

Nvm... The language report has a specification error...

joshuabach:
I've pushed some of my updates to the parser2 branch. Can't rebase on this, since it doesn't compile... Think before pushing @reiner-dolp!

reiner-dolp:
I was totally aware what I was doing.

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+
+impl ExpectedToken for BinaryOp {
+    type Yields = (Operator, Precedence, Assoc);
+    fn matching(&self, token: &TokenKind) -> Option<Self::Yields> {
+        match token {
+            // TODO Linear search. meh.
+            TokenKind::Operator(op) => BINARY_OPERATORS
+                .iter()
+                .find(|(this_op, _, _)| this_op == op)
+                .map(|elt| *elt),
```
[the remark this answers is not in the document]
joshuabach: short for `element`. Kinda like `itm` for `item`

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+        }
+
+        Ok(v)
+    }
+
+    /// In average compilers this is sometimes called `expect`
+    fn omnomnom<E, G>(&mut self, want: G) -> SyntaxResult<'f, Spanned<'f, E::Yields>>
+    where
+        E: ExpectedToken,
+        G: Into<E>,
```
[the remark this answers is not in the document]
joshuabach: resolving for PR-clarity, since no change requested ;-)

src/main.rs
joshuabach: Yes.

src/parser.rs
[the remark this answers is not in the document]
flip1995: both shouldn't be abbreviated.

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+        while v.len() < n {
+            let eof_token = self.eof_token.as_ref().ok_or(SyntaxError::MissingEOF)?;
+            v.push(eof_token)
+        }
+
+        Ok(v)
+    }
+
+    /// In average compilers this is sometimes called `expect`
+    fn omnomnom<E, G>(&mut self, want: G) -> SyntaxResult<'f, Spanned<'f, E::Yields>>
```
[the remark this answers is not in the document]
flip1995: Let's say we agree to disagree. But I'm happy both ways, as long as it works.

Still: please rename these functions to `eat` and `try_eat` or  something similar. `omnomnom` is just a step to far.

grammer.md
joshuabach: I now did a psuedo-ast to explicitly check for this, which we should keep until we get a real AST. I discussed this with @problame. We considered it more elegant to treat MainMethod as special Methods. Same for NewObjectExpression. See 0dc442f

joshuabach:
17/90 mjtests are currently not passing. The highest priority should be fixing these, and than I suggest we merge this. Anything else (diagnostics integration, cleanup, var renaming, etc) can be addressed in seperate PRs. This one is already to massive.

joshuabach:
@flip1995 Why did you put the dangling `else` under `Required`? Technically, we can defer this until the AST generation phase next week

flip1995:
Dangling else can be postponed. But currently we don't handle `else` at all.

src/parser.rs
```diff
@@ -0,0 +1,831 @@
...
+            .is_some()
+        {
+            // empty statement
+            Ok(())
+        } else if self.omnomnoptional::<Exactly, _>(Keyword::If)?.is_some() {
+            self.omnomnom::<Exactly, _>(Operator::LeftParen)?;
+            self.parse_expression()?;
+            self.omnomnom::<Exactly, _>(Operator::RightParen)?;
+
+            self.parse_statement()
```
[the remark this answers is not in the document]
joshuabach: Oops. forgot ;-)

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+}
+const BINARY_OPERATORS: &[(Operator, Precedence, Assoc)] = &[];
+
+#[derive(Debug, Clone)]
+pub enum SyntaxError<'f> {
+    // TODO Rather panic? If `MissingEOF` is instantiated, we have most certainly a bug
+    MissingEOF,
+    UnexpectedToken {
+        got: Token<'f>,
+        expected: String, // TODO This is temporary, shouldn't be string
```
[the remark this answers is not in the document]
joshuabach: @reiner-dolp will update in seperate PR

src/parser.rs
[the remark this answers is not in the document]
joshuabach: I agreed with @flip1995  on `nom` and `nom_optional` in real-world discussion. Will do in separate PR. see #30

src/parser.rs
```diff
@@ -0,0 +1,785 @@
...
+    fn tastes_like<E, G>(&mut self, want: G) -> SyntaxResult<'f, bool>
+    where
+        E: ExpectedToken,
+        G: Into<E>,
+    {
+        self.peek().map(|got| want.into().matches(&got.data))
+    }
+
+    fn parse_program(&mut self) -> ParserResult<'f> {
+        while self.omnomnoptional::<EOF, _>(EOF)?.is_none() {
```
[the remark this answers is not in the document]
joshuabach: Maybe if we are bored / motivated ;-)

PR #35, opened 2018-11-03
render parser errors using the diagnostics interface

reiner-dolp:
image
image
image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@
 #[macro_use]
 extern crate derive_more;
 
+use self::diagnostics::MaybeSpanned;
 use failure::{Error, Fail, ResultExt};
 use memmap::Mmap;
 use std::{fs::File, io, path::PathBuf, process::exit};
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -192,9 +191,18 @@ fn cmd_parsetest(path: &PathBuf) -> Result<(), Error> {
...
     match parser.parse() {
         Ok(_) => Ok(()),
-        // TODO: generate error message per spec
-        Err(err) => {
-            eprintln!("error: {}", err);
+        Err(parser_error) => {
+            // TODO: context.error should do this match automatically through
+            // generic arguments
+            match parser_error {
+                MaybeSpanned::WithSpan(spanned) => context.error(Spanned {
```
joshuabach:
```suggestion
                MaybeSpanned::WithSpan(spanned) => spanned.map(Box::new)
```

joshuabach: This also replaces the next 3 lines

src/parser.rs
```diff
@@ -207,7 +198,7 @@ where
         self.peek()?;
         match self.lexer.next() {
             Some(token) => Ok(token),
-            None => self.eof_token.clone().ok_or(SyntaxError::MissingEOF),
+            None => self.eof_token.clone().ok_or(MaybeSpanned::WithoutSpan(SyntaxError::MissingEOF)),
```
joshuabach:
```rust
use MaybeSpanned::*;
```

src/main.rs
reiner-dolp: This suggestion does not work.

PR #28, opened 2018-11-01
LibFirm bindings + infra for wrapper crate

problame:

PR #37, opened 2018-11-03
mjtest crate: procedural macros for re-using mjtest cases as cargo tests

problame:


reiner-dolp:
There are also some clippy warnings that need fixing

problame:
`cargo fmt` has no effect on my machine, though

reiner-dolp:
bors retry

reiner-dolp:
We also have a collition for the mj-test CI jobs

```
fatal: destination path 'mjtest' already exists and is not an empty directory.
```

mjtest-rs/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "mjtest"
+version = "0.0.1"
+authors = [
+    "flip1995 <hello@philkrones.com>",
+    "hediet <>",
+    "joshuabach <>",
+    "problame <>",
+    "reiner-dolp <>"
+]
+# edition = "2018" https://users.rust-lang.org/t/how-to-use-proc-macro-on-rust-2018/20833/3
```
reiner-dolp: Is this garbage or intentional?

reiner-dolp:
I am not sure, but maybe we have to add a line to `build` that clones the submodules? The CI does it correctly, but I am not sure about the test infrastructure of the IPD.

```
git submodule update --init --recursive
```

mjtest-rs/Cargo.toml
problame: intentional, if someone wonders why we can't use 2018 yet

problame:
>     * I am not sure, but maybe we have to add a line to `build` that clones the submodules? The CI does it correctly, but I am not sure about the test infrastructure of the IPD.
>       ```
>       git submodule update --init --recursive
>       ```

`./build` does not invoke `cargo test`, so nothing fails in speedcenter

reiner-dolp:
Ah, right :) but then we cannot use the build script locally for setup during devevlopment. In this case I just have to remember.

reiner-dolp:
The CI fails because you cannot pass `--release` to `cargo fmt`

problame:
> Ah, right :) but then we cannot use the build script locally for setup during devevlopment. In this case I just have to remember.

I have a re-write of the README.md in https://github.com/comprakt/comprakt/pull/28/commits/fcf46bdda5aa798d492aa6af15f11f07fb6ae37e which will cover `git submdoule update --init --recursive`.

.travis.yml
```diff
@@ -12,48 +12,73 @@ branches:
     # This is enabled to build pull requests.
     - master
 
+env:
+  - TEST_KIND="mjtest-lexer"
```
flip1995: Why do we need this? Everywhere in the matrix the `TEST_KIND` is defined. This shouldn't be required.

.travis.yml
```diff
@@ -12,48 +14,73 @@ branches:
...
-      MJ_TIMEOUT=60 MJ_RUN=target/release/comprakt python3 ./mjtest/mjt.py $MJ_TESTS
-    fi
+  - > 
+    case "$TEST_KIND" in
+      comprakt-*)
+        case "$TEST_KIND" in
+          comprakt-release) FLAGS="--release"
+        esac
+        cargo fmt --all -- --check && \
+        cargo clippy $FLAGS --all-targets --all-features -- -D warnings && \
```
flip1995: We also don't need a `--release` flag on Clippy.

reiner-dolp:
looks good to me now. Great work!

PR #46, opened 2018-11-09
Refactor positions and spans

reiner-dolp:
Makes all constructors for spans and positions private. New positions and spans have to be generated by combination of safe conversions.

Also see notes on commit `remove EOF token` and `[refactor] position,span,asciifile to own files; remove PositionedChar`.


reiner-dolp:
##TODO
- [ ] Check performance implications, I suspect a slow down.
- [ ] Reimplement Diagnostics output.

reiner-dolp:
I haven't optimized anything yet.

This branch:
```
rd ~/r/comprakt> perf stat -- ./target/release/comprakt --lextest mjtest-tests/lexer/huge_file.mj > /dev/null

 Performance counter stats for './target/release/comprakt --lextest mjtest-tests/lexer/huge_file.mj':

       8066,354664      task-clock:u (msec)       #    1,000 CPUs utilized          
                 0      context-switches:u        #    0,000 K/sec                  
                 0      cpu-migrations:u          #    0,000 K/sec                  
             1.129      page-faults:u             #    0,140 K/sec                  
    14.128.593.069      cycles:u                  #    1,752 GHz                    
    34.539.193.258      instructions:u            #    2,44  insn per cycle         
     5.387.559.462      branches:u                #  667,905 M/sec                  
         8.260.726      branch-misses:u           #    0,15% of all branches        

       8,066785787 seconds time elapsed

       6,003629000 seconds user
       2,053508000 seconds sys
```
Current master:
```
rd ~/r/comprakt> perf stat -- ./target/release/comprakt --lextest mjtest-tests/lexer/huge_file.mj > /dev/null

 Performance counter stats for './target/release/comprakt --lextest mjtest-tests/lexer/huge_file.mj':

      11364,823887      task-clock:u (msec)       #    0,999 CPUs utilized          
                 0      context-switches:u        #    0,000 K/sec                  
                 0      cpu-migrations:u          #    0,000 K/sec                  
             1.130      page-faults:u             #    0,099 K/sec                  
    22.854.264.562      cycles:u                  #    2,011 GHz                    
    68.979.552.843      instructions:u            #    3,02  insn per cycle         
    12.532.006.443      branches:u                # 1102,701 M/sec                  
        17.411.736      branch-misses:u           #    0,14% of all branches        

      11,380485004 seconds time elapsed

       9,266102000 seconds user
       2,075789000 seconds sys
```

src/diagnostics.rs
```diff
@@ -24,6 +19,14 @@ pub struct Diagnostics {
     writer: RefCell<Box<dyn WriteColor>>,
 }
 
+const ENCODING_ERROR_MAX_CONTEXT_LENGTH: usize = 80;
```
problame: my usual nitpick

src/diagnostics.rs
```diff
@@ -233,138 +236,136 @@ impl Message {
...
-                    }
-                }
-            }
-
-            // add padding line below
-            output.set_color(HIGHLIGHT);
-            output.set_bold(true);
-            writeln!(output.writer(), "{}", empty_line_marker);
-        }
+        //let line_number_width = (span.end.row + 1).to_string().len();
```
problame: comment commited

src/diagnostics.rs
reiner-dolp: sorry. I forgot the WIP on the pull request name. diagnostics.rs is not yet rewritten. Most of the file is just commented out to make it compile.

reiner-dolp:
Yes! It's a really nice side effect of using spans. As the comments in the code say. Span is just a "fat" version of a slice with a 2D column-row position instead of just a byte offset. Same goes for Position, which is just a 2D index instead of a "normal array offset".

But I think there is a lot to optimize. Haven't looked into performance yet.

src/diagnostics.rs
reiner-dolp: see comment above. This is temp code to make it compile until the code is rewritten.

asciifile/src/span.rs
```diff
@@ -0,0 +1,286 @@
...
+//! A `Span` is immutable by design. Mutations will always return a new
+//! instance of a Span.
+//!
+//! A `Span` is a guranteed valid text range of the input file. This means
+//! actions like `as_str()` cannot fail with out of bounds indices on run-time.
+//!
+//! The range defined by the start and end position of a `Span` is inclusive on
+//! the both sides (lower and upper bound). As a result, `as_str()` will contain
+//! the character pointed at by the start position and the character pointed to
+//! by the end position. Beware that this is not equivalent to rusts behaviour
```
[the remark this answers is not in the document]
reiner-dolp: It's not plural, it's the possessive "the behaviour of rust"?

reiner-dolp:
Empty spans not being representable is by design. Otherwise a lot of functions would be wrapped in Option<> just to deal with an empty file. e.g. the obvious behaviour being position.byte(), but also more weird behaviour like span.start_position().chr() would have to behave differently than span.get_first_chr()...

reiner-dolp:
The easiest fix would be to make parse()/parse_program() return a `ParserResult<'f, ParsedProgram>` value of
```
enum ParsedProgram<'f> {
    EmptyFile,
    NonEmptyFile(ast::Program<'f>)
}
```
Just filter it directly in parse(). The semantic analysis phase can reject the empty file with a "missing main" immediatly when it recieves the `ParsedProgram`.

joshuabach:
Hm. In my heart, I believe there exists a design that uses exclusive-spans, a `Spanned<EOF>` token and a never-ending iterator that works without special cases. But maybe that's an illusion... I'll do it your way

joshuabach:
Done. Once all tests pass maybe you should take a look at it (you know the span/pos code better than me), otherwise I think we can merge this

asciifile/src/span.rs
```diff
@@ -0,0 +1,286 @@
...
+    ///     .iter()
+    ///     .take_while(|position| position.chr().is_lowercase())
+    ///     .collect::<Vec<_>>();
+    ///
+    /// let span = Span::from_positions(&positions).unwrap();
+    /// assert_eq!("abcdfegh", span.as_str());
+    /// ```
+    pub fn as_str(&'f self) -> &'f str {
+        // the range is inclusive on both sides!
+        &self.start.file()[self.start.byte_offset()..=self.end.byte_offset()]
```
reiner-dolp: Array indexing on Stringtable instead of u8 array. I'll fix this during merge.

src/diagnostics.rs
```diff
@@ -166,283 +184,223 @@ pub struct Message {
...
+                        output.writer(),
+                        "{spaces}{underline}",
+                        spaces = " ".repeat(start_term_pos),
+                        underline = "^".repeat(term_width)
+                    )?;
+                }
 
-    fn writer(&mut self) -> &mut dyn WriteColor {
-        self.writer
+                // debug
```
reiner-dolp: Remove

problame:
I checked the memory footprint and runtime impact of this change using `perf stat`: nothing significant, but positive improvements in runtime, less total number of branches, etc.

PR #47, opened 2018-11-09
patternmatching-based visitor + sample use: AST-dumper

problame:
based on `ast-integration branch`


problame:
Upfront: the macros could be refactored into procmacros, but it's much easier to figure out what is going on / wrong when using normal macros (compiler warnings are also way better!)

PR #48, opened 2018-11-09
Switch from white listing branches to black listing them

flip1995:
r? @reiner-dolp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,11 @@ rust:
 cache: cargo
 
 branches:
-  only:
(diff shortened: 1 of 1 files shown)


flip1995:
The bas thing about this is that every PR will now get checked twice, if not created from a fork.

reiner-dolp:
Once without merge and once against a merge with master, right?

flip1995:
Yes. The `travis-ci/pr` check is with a merge against the master and the `travis-ci/push` check is just the new branch.

PR #49, opened 2018-11-10
Add attributed grammar and abstract syntax

joshuabach:


reiner-dolp:
Shouldn't both be moved into the tasksheets/sheet2/ subfolder?

joshuabach:
At least the `grammer.md` is not only for any specific assignment, but also a documentation of our code. As it (fairly) resembles what is implemented in parser.rs

problame:
bors try

problame:
bors r=flip1995,problame

PR #50, opened 2018-11-10
Implement pretty printing

hediet:


src/lexer.rs
```diff
@@ -85,6 +85,10 @@ impl<'f, T> Spanned<'f, T> {
             data: f(&self.data),
         }
     }
+
+    pub fn get_data(&self) -> &T {
+        &self.data
+    }
```
joshuabach: Spanned implements `Deref`, so this is unnecassary. Instead of doing `spanned.get_data()`, you can just write `*spanned` (Btw, `data` is a public member anyway, so this getter is double-unnecesarry)

src/prettyprint.rs
```diff
@@ -1,10 +1,398 @@
...
-    println!("way to go:\n{:#?}", program);
+
+    let mut stdout = std::io::stdout();
+    let mut printer = IndentPrinter::new(&mut stdout);
+    do_prettyprint(&NodeKind::from(program), &mut printer);
     Ok(())
 }
+
+#[allow(clippy::cyclomatic_complexity)]
+fn do_prettyprint(n: &NodeKind<'_, '_>, printer: &mut IndentPrinter<'_>) {
```
joshuabach: Why not a member function on the printer?

src/prettyprint.rs
```diff
@@ -1,10 +1,398 @@
...
+                Method(ty, params, block) => {
+                    printer.print_str(&"public ");
+                    do_prettyprint(&NodeKind::from(&ty.data), printer);
+                    printer.print(format_args!(" {}(", member.name));
+                    do_prettyprint(&NodeKind::from(&params.data), printer);
+                    printer.print_str(&") ");
+                    do_prettyprint(&NodeKind::from(&block.data), printer);
+                }
+                MainMethod(param_name, block) => {
+                    printer.print(format_args!("public static void main(String[] {})) ", param_name));
```
joshuabach:
```suggestion
                    printer.print(format_args!("public static void {}(String[] {})) ", member.name, param_name));
```

src/prettyprint.rs
```diff
@@ -1,10 +1,398 @@
...
+                }
+            }
+        }
+        Type(ty) => {
+            do_prettyprint(&NodeKind::from(&ty.basic), printer);
+            printer.print(format_args!("{}", "[]".repeat(ty.array_depth as usize)));
+        }
+        BasicType(basic_ty) => {
+            use crate::ast::BasicType::*;
+            let _ = match basic_ty {
```
joshuabach: What's the `let _` for?

src/prettyprint.rs
```diff
@@ -1,10 +1,398 @@
...
+    let bkind = ast::ClassMemberKindDiscriminants::from(&b.kind);
+    match (akind, bkind) {
+        (Field, Field)
+        | (Method, Method)
+        | (MainMethod, MainMethod)
+        | (Method, MainMethod)
+        | (MainMethod, Method) => a.name.cmp(&b.name),
+        (Method, Field) | (MainMethod, Field) => std::cmp::Ordering::Less,
+        (Field, Method) | (Field, MainMethod) => std::cmp::Ordering::Greater,
+    }
```
joshuabach:
```rust
    match (akind, bkind) {
        (Method, Field) | (MainMethod, Field) => std::cmp::Ordering::Less,
        (Field, Method) | (Field, MainMethod) => std::cmp::Ordering::Greater,
        (_, _) => a.name.cmp(&b.name),
    }
```

src/prettyprint.rs
hediet: The printer is a standalone, reusable component. Why mixing things?

problame:
I am rebasing this branch onto https://github.com/comprakt/comprakt/pull/53

problame:
Please review and admit https://github.com/comprakt/comprakt/pull/53 first

src/prettyprint.rs
joshuabach: It just seems weird having a bunch top-level functions and passing the same object to each and every one of it. Maybe define a `struct PrettyPrinter(IndentPrinter<'_>)` and make the member functions of those. But not today.

src/prettyprint.rs
flip1995: Always be explicit if possible!

src/main.rs
```diff
@@ -83,11 +87,16 @@ enum CliCommand {
         #[structopt(name = "FILE", parse(from_os_str))]
         path: PathBuf,
     },
-    #[structopt(name = "--printast")]
+    #[structopt(name = "--print-ast")]
```
problame: this addresses complaints in https://github.com/comprakt/comprakt/issues/57

src/main.rs
```diff
@@ -83,11 +87,16 @@ enum CliCommand {
...
     PrintAst {
         #[structopt(name = "FILE", parse(from_os_str))]
         path: PathBuf,
     },
+    #[structopt(name = "--debug-dumpast")] // nonstandard
```
problame: this addresses complaints in https://github.com/comprakt/comprakt/issues/57

tests/integration.rs
```diff
@@ -19,6 +21,8 @@ fn compiler_flag(phase: CompilerPhase) -> &'static str {
     match phase {
         CompilerPhase::Lexer => "--lextest",
         CompilerPhase::Parser => "--parsetest",
+        CompilerPhase::AstStructurePrint => "--print-ast",
+        CompilerPhase::AstPrettyPrint => "--debug-dumpast",
```
problame: `--debug-dumpast` was not meant to be a stable interface... the tests leak the implementation detail of AST node type names, etc.

flip1995:
I'm currently fixing some stuff in the PP

tests/integration.rs
reiner-dolp: ok, ill remove it.

src/prettyprint.rs
```diff
@@ -217,12 +216,13 @@ fn do_prettyprint(n: &NodeKind<'_, '_>, printer: &mut IndentPrinter<'_>) {
                         match els.data {
                             Empty => (),
                             _ => {
-                                printer.print_str(&"else ");
```
joshuabach: if you do this, you can also collapse the nested `match`

joshuabach: nvm

src/lexer.rs
reiner-dolp: I cannot find it. Seems to be resolved.

src/prettyprint.rs
reiner-dolp: I am voting for not a member function but its own module. `print::pretty::format` oder something like that.

src/prettyprint.rs
reiner-dolp: Seems to be gone.

src/prettyprint.rs
reiner-dolp: Seems more like a debate. I am pro explicit.
Marking as resolved.

integration-tests/ast/if_else.mj.stdout
```diff
@@ -0,0 +1,13 @@
+class IfElse {
+	public static void main(String[] args) {
+		if (true)
+			return;
```
problame: why no braces around the `return` ( as well as in line 7 ?

flip1995: This is correct. In Java the braces are optional if the `if` body consists of a single statement.

Or is there a specification, that says that `if` bodies always have to be parenthesized?

src/prettyprint.rs
problame: I'll take care of that.

problame: done in 9086d03edebe2236acb2db1fb3b21dc251a74e28

problame:
bors try

integration-tests/ast/if_else.mj.stdout
problame: I just wonder how we distinguish this one comparated to the two single-statement `if` / `else` variants in the `hello_world.mj`. WIll make a comment there...

integration-tests/ast/hello_world.mj.stdout
```diff
@@ -0,0 +1,19 @@
...
+	public static void main(String[] args) {
+		(System.out).println(43110 + 0);
+		boolean b = true && (!false);
+		if ((23 + 19) == ((42 + 0) * 1))
+			b = (0 < 1);
+		else if (!(array[2 + 2])) {
+			int x = 0;
+			x = (x + 1);
+		} else {
+			(new HelloWorld()).bar(42 + (0 * 1), -1);
```
problame: how is this single-statement `else` block different from the single-statement `if` block above (line 8--10 in this file)

reiner-dolp: So, is this wrong or correct? I am getting really confused now. Especially since the txt files on the compilerpraktikum website are obviously not correct too

flip1995: In the input file were braces, so the pretty printer will also print braces (`block` statement)

PR #53, opened 2018-11-11
implement visitor based on pattern matching

problame:
this PR is the basis for the prettyprinting PR https://github.com/comprakt/comprakt/pull/50


joshuabach:
This seems to have been tested (at least on basically) by the prettyprinter in #50. So I guess merge?

reiner-dolp:
- [ ] the cli flag is incorrectly named, should be "--print-ast". But that is fixed by the pretty printer branch
- [ ] structureprint.rs and prettyprint.rs not being in a subfolder "print/{structure, pretty}.rs" seems weird. Namespaces would make more sense to: `print::structure::Printer` and `print::pretty::Printer`.

reiner-dolp:
See my notes above, but I would like to go ahead and merge it anways.

PR #72, opened 2018-11-12
CLI infrastructure for semantic analysis compiler stage

problame:


src/sem.rs
```diff
@@ -0,0 +1,57 @@
...
+use failure::{format_err, Error, Fail};
+use std::{collections::HashMap, rc::Rc};
+
+type MethodDefinitions<'a, 'f> = HashMap<
+    (Symbol, Symbol, ast::ClassMemberKindDiscriminants),
+    &'a Spanned<'f, ast::ClassMember<'f>>,
+>;
+
+pub fn check<'a, 'f>(_ast: &'a ast::AST<'f>, context: &context::Context<'_>) -> Result<(), Error> {
+    let mut member_definitions: MethodDefinitions<'a, 'f> = HashMap::new();
```
flip1995: I would write visitors which need to store some information as:

```rust
struct WhatAmIDoingVisitor {
    member_defs: MethodDefinitions,
}

impl WhatAmIDoingVisitor {
    first_pass(&mut self, n: NodeKind, cx: Context) {
        ...
    }
}
```

problame: I will make a separate PR with the POC visitor. This branch got out of hand, will force push something that fits the title of this PR.

PR #77, opened 2018-11-12
Make most of the NodeKinds of the visitor Spanned

flip1995:


problame:
I did some work on top of this PR, things seemed to work just fine.
bors r+

PR #78, opened 2018-11-12
fix #74: take dyn Display instead of dyn AsFail in Diagnostics Api

### LINKED ISSUE #74: Make Diagnostics API take Display instead of AsFail
At least info(). It evolved to something it was not intended to be.

reiner-dolp:
Fail has Display as Trait bound, and we only use the Display API, not any of the Fail APIs. Since Diagnostics was extended to writing info messages in addition to error messages, changing the trait bound does not only make sense from a technical, but also a semantic view point.

We can now call Diagnostics as follows:

```rust
ctx.diagnostics.info(&Spanned {
    span,
    data: "span for AST node 'whole program with trimmed whitespace'"
});
```
to generate
```
info: span for AST node 'whole program with trimmed whitespace'
   | 
 2 |             class Foo {
   |             ^^^^^^^^^^^
 3 |                 public static void main(String[] args) {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 4 |                     return x || y = z  ;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 5 |                 }
   | ^^^^^^^^^^^^^^^^^
 6 |             }
   | ^^^^^^^^^^^^^

```

Or without a span:
```rust
ctx.diagnostics.info(&WithoutSpan("bla bla"));
ctx.diagnostics.info(&"test");
```


reiner-dolp:
Let me just clean up a little bit.

reiner-dolp:
@problame this can be reviewed.

reiner-dolp:
I am merging this. Otherwise I cannot continue working. We can revert or fix things later if someone disagrees.

PR #79, opened 2018-11-12
visitor: for_each_child with VisitResult return type

problame:


src/visitor.rs
```diff
@@ -113,25 +148,25 @@ impl<'a, 't> NodeKind<'a, 't> {
                 match &e.data {
                     Binary(_, lhs, rhs) => {
                         ccb!(lhs.as_ref());
-                        ccb!(rhs.as_ref());
+                        ccb!(rhs.as_ref())
```
flip1995: If there are 2 following `ccb`s, shouldn't it return already if the first fails?

src/visitor.rs
```diff
@@ -47,64 +47,99 @@ impl Debug for NodeKind<'_, '_> {
...
+    {
         use self::NodeKind::*;
         macro_rules! ccb {
-            ($astvar:expr) => {
-                cb(NodeKind::from($astvar))
-            };
+            ($astvar:expr) => {{
+                let res = cb(NodeKind::from($astvar));
+                if res.stop_visit() {
+                    return Some(res);
```
flip1995: ?
```suggestion
                    return None;
```

problame: No, as described in the (misplaced) doc comment above, `for_each_child` returns None if no children were visited, otherwise the return value of the callback.

src/visitor.rs
```diff
@@ -47,64 +47,99 @@ impl Debug for NodeKind<'_, '_> {
...
+
+impl<X, E> VisitResult for Result<X, E> {
+    fn stop_visit(&self) -> bool {
+        self.is_err()
+    }
+}
+
+/// Visit the children of `self`, invoking `cb` on each.
+/// If `self` has no children, this method returns `None.
+/// Otherwise, it returns `Some(res)` where `res` is the return value of `cb`.
```
problame: misplaced, should be above `for_each_child`

src/visitor.rs
problame: Doesn't the macro just do that?

PR #80, opened 2018-11-12
disable travis cache

reiner-dolp:


reiner-dolp:
ok. actually slower because strum_macros and derive_more take ages to compile.

PR #83, opened 2018-11-13
turn project into virtual workspace

reiner-dolp:
- clearer distinction between compiler library and cli interface
- more granular assignment of dependencies to crates
- more clear which tests run on which infrastructure (interation tests are part of compiler-cli, mj-tests are part of compiler-lib)
- allows for doc tests
- move asciifile back into compiler library
- move lexer tests in old main into lexer module


flip1995:
1 commit - 355 files changed. Oh boi let's review :tada:

compiler-lib/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+    "hediet <>",
+    "joshuabach <>",
+    "problame <>",
+    "reiner-dolp <>"
+]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+structopt = "0.2"
```
flip1995: Do we need structopt in compiler-lib?

reiner-dolp: Nope. :)

Cargo.toml
```diff
@@ -1,34 +1,19 @@
...
-memmap = "0.7.0"
-libc = "*"
-termcolor = "1.0.4"
-derive_more = "0.13"
-asciifile = { path = "./asciifile" }
-strum = "0.11.0"
-strum_macros = "0.11.0"
+members = [
+    "compiler-lib",
+    "compiler-cli",
```
reiner-dolp: @flip1995 The only side-effect of listing crates seems to be a shared cargo.lock. Should we also list the integration test crates in compiler-cli?

flip1995: It would at least be more consistent to list them here, since we're also listing  `mjtest-rs/mjtest_macros` here.

reiner-dolp: or remove mjtest from the list

PR #85, opened 2018-11-13
visitor: for_each_child with VisitResult return type

problame:
This commit has already been reviewed in https://github.com/comprakt/comprakt/pull/79 but fell under the bors bus


reiner-dolp:
did I accidently delete the patch or was there a bug in bors?

reiner-dolp:
btw. This module could use some docs with an example :)

PR #86, opened 2018-11-13
sem: first-pass visitor + scope types + MainMethod with paramlist

problame:
A first draft of how the semantic checkers could look like.

* Scope is quite inefficient ans more or less a placeholder (it doesn't even support lookups)
  * But it servers as a demo for different scope kinds
* the re-definition error (see `format_err`) should be done properly. maybe @reiner-dolp can explain how we'd do this? do we even need a Result if we just use Context to highlight sutff?


compiler-lib/src/sem.rs
```diff
@@ -1,6 +1,163 @@
...
+    pub fn visit<'a, 'f>(
+        ast: &'a ast::AST<'f>,
+        context: &context::Context<'_>,
+        _classes_and_members: &ClassesAndMembers<'a, 'f>,
+    ) -> Result<(), Error> {
+        // TODO java.lang.System.out scope
+        // TODO ontop of that, scope of this package
+        let mut vardefs: Rc<Scope<VarDef<'a, 'f>>> = Scope::root();
+        let mut tydefs: Rc<Scope<TyDef>> = Scope::root();
+        let mut methdefs: Rc<Scope<MethDef>> = Scope::root();
```
flip1995: Put all this data structures in the `SecondPass` struct, make the `visit` method a `new` function.

Then call 
```rust
let visitor = SecondPass::new(cx)
visitor.do_visit(ast)
```
from the check function.

Same goes for FirstPass and should be our standard approach on implementing visitors IMO. That way a visitor doesn't need as many parameters as data it has to hold.

compiler-lib/src/sem.rs
```diff
@@ -1,6 +1,163 @@
...
+        let key = Rc::clone(sym);
+        match self.definitions.entry(key) {
+            Entry::Occupied(o) => return Err(()),
+            Entry::Vacant(e) => e.insert(val),
+        };
+        Ok(())
+    }
+}
+
+struct SecondPass;
```
reiner-dolp: I would love some comments here. E.g. It would be perfect to list what the "second pass" should do in a doc comment. Same for FirstPass.

compiler-lib/src/parser.rs
```diff
@@ -332,17 +332,36 @@ where
                 // Consume exactly `void IDENT(String[] IDENT)`
                 self.omnomnom(exactly(Keyword::Void))?;
                 let name = self.omnomnom(Identifier)?.data;
-                self.omnomnom(exactly(Operator::LeftParen))?;
-                self.omnomnom(exactly(Symbol::from("String")))?;
+                let pl_begin = self.omnomnom(exactly(Operator::LeftParen))?;
```
reiner-dolp: pl => left paren

compiler-lib/src/sem.rs
```diff
@@ -1,6 +1,163 @@
...
+                        res.insert(
+                            (Rc::clone(&class.name), Rc::clone(&member.name), discr),
+                            &member,
+                        );
+                    }
+                }
+                Ok(())
+            }
+            x => {
+                return Err(format_err!(
```
reiner-dolp: why not tighten the type definition of do_visit to ast::Ast::Program() to get rid of this impossible state?

reiner-dolp: Or the only other possible state seems to be `ast::AST::Empty`, why not just reject the program at this point with "missing main"?

compiler-lib/src/sem.rs
```diff
@@ -1,6 +1,163 @@
...
+        match n {
+            ClassMember(m) if m.kind.is_method() => {
+                // Assume types have been checked
+                // TODO enforce somewhere that main has String[] arg type
+                let params = m.kind.method_params().unwrap();
+                let mut vardefs = Scope::enter_scope(vardefs);
+                for p in &params.data {
+                    vardefs
+                        .define(&p.name, &p.ty)
+                        .or(Err(format_err!("re-definition of variable in same scope")))?;
```
reiner-dolp: what's the problem here??? you can generate a normal derive(Fail) error here rendering the second definition using p.span.
Or do you want to attach 2 spans pointing at both definitions? We could implement multispans.

reiner-dolp:
Could we also use a longer name like "semantic_analysis.rs" or "semantics.rs" instead of "sem.rs"?

compiler-lib/src/sem.rs
reiner-dolp: >  we even need a Result if we just use Context to highlight sutff?

No. You just need anything that can be a string or has Display implemented. You could just do:
```
context.diagnostics.error(Spanned {
    span: p.span,
    data: SemanticError::RedefinitionError {
         kind: "variable", name: p.name.to_string()
    }
})
```
or: 
```
context.diagnostics.error(Spanned {
    span: p.span,
    data: "re-definition of variable in same scope"
})
```

compiler-lib/src/parser.rs
reiner-dolp: ah wait! pl = parmeter list, but this makes my argument! the name is too short.

compiler-lib/src/parser.rs
```diff
@@ -332,17 +332,36 @@ where
...
+                let java_string = self.omnomnom(exactly(Symbol::from("String")))?;
                 self.omnomnom(exactly(Operator::LeftBracket))?;
-                self.omnomnom(exactly(Operator::RightBracket))?;
-                let param = self.omnomnom(Identifier)?.data;
-                self.omnomnom(exactly(Operator::RightParen))?;
+                let java_string_array_close = self.omnomnom(exactly(Operator::RightBracket))?;
+                use crate::asciifile::{Span, Spanned};
```
reiner-dolp: move to top of file

compiler-lib/src/sem.rs
```diff
@@ -1,6 +1,163 @@
...
+    fn do_visit<'a, 'f>(
+        n: &NodeKind<'a, 'f>,
+        _context: &context::Context<'_>,
+        res: &mut ClassesAndMembers<'a, 'f>,
+    ) -> Result<(), Error> {
+        use self::{ast, NodeKind::*};
+        match n {
+            AST(ast::AST::Program(p)) => {
+                for class in &p.data.classes {
+                    for member in class.members.iter() {
```
flip1995: NIT: [`explicit_iter_loop`](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#explicit_iter_loop) inconsistent usage here

compiler-lib/src/sem.rs
flip1995: There could be a lot of other states. `NodeKind` could be any `AST`-node. So if you call this function with an `Expr` it will error with
```
"unexpected ast node passed to first pass: Expr",
```

reiner-dolp: I don't get it. He is not using recursion at all, since all ClassDefintions are at the top level. So why make the method parameter a AST-Node and not the Root Node Program/Empty which avoids this issue?

flip1995: Yeah this doesn't really need a visitor in that case then.

compiler-lib/src/ast.rs
```diff
@@ -48,7 +48,7 @@ pub enum ClassMemberKind<'t> {
         Spanned<'t, ParameterList<'t>>,
         Spanned<'t, Block<'t>>,
     ),
-    MainMethod(Symbol, Spanned<'t, Block<'t>>),
+    MainMethod(Spanned<'t, ParameterList<'t>>, Spanned<'t, Block<'t>>),
```
joshuabach: Why? The reason I made this a `Symbol` was to avoid  having to add a documented invariant stating that the list is "guaranteed to only contain [...]". Below, `method_params` should return `None` for `MainMethod`, since for all our intents and purposes the MainMethod has no parameter. It just has a symbol that must not be used.

compiler-lib/src/ast.rs
```diff
@@ -201,3 +201,30 @@ pub enum UnaryOp {
...
+            Method(_, _, _) => true,
+            MainMethod(_, _) => true,
+            Field(_) => false,
+        }
+    }
+    pub fn method_params(&self) -> Option<&Spanned<'_, ParameterList<'_>>> {
+        use self::ClassMemberKind::*;
+        match &self {
+            Method(_t, pl, _block) => Some(pl),
+            MainMethod(pl, _block) => Some(pl),
```
joshuabach:
```suggestion
            MainMethod(_, _, _) => None,
```

compiler-lib/src/semantics.rs
```diff
@@ -0,0 +1,162 @@
...
+        context: &context::Context<'_>,
+        vardefs: Rc<Scope<VarDef<'a, 'f>>>,
+        tydefs: Rc<Scope<TyDef>>,
+        methdefs: Rc<Scope<MethDef>>,
+    ) -> Result<(), Error> {
+        use self::NodeKind::*;
+        match n {
+            ClassMember(m) if m.kind.is_method() => {
+                // Assume types have been checked
+                // TODO enforce somewhere that main has String[] arg type
```
joshuabach: We wouldn't have to enforce this, if the `MainMethod` had just a `Symbol` and not `ParameterList`...

compiler-lib/src/semantics.rs
```diff
@@ -0,0 +1,162 @@
...
+                let params = m.kind.method_params().unwrap();
+                let mut vardefs = Scope::enter_scope(vardefs);
+                for p in &params.data {
+                    vardefs
+                        .define(&p.name, &p.ty)
```
joshuabach: To realize this whitout have a full parameter list on `MainMethod`, make `VarDef` an enum of `Reserved` and `Binding(ast::Type)` and than treat main method specially here, adding a `vardefs.reserve(&mainmethod_symbol)`. I like that much better than carrying around the "main method params are a list of length 1 with the only parameter being an array of `String`" invariant

compiler-lib/src/ast.rs
flip1995: The only reason why this would make sense is that we have a span around the whole parameter. It is already enforced by the parser that the type of the argument is `String[]` and that the `MainMethod` only has one parameter.

problame: We need `args` for the local variable declaration scope in when doing name analysis. AFAIK it must not be shadowed by the body of the `main()` function unless main opens a new scope. but the body of `main` and the param list are in the same scope.

compiler-lib/src/ast.rs
problame: due to type inference, the variable name is the only way to catch changed semantics (e.g. different ordering of struct members) during code review. but then again, the order-changing should have not passed code review anyways...

compiler-lib/src/semantics.rs
problame: that TODO is legacy. again, we need the variable name for the scope of the method body, so it makes sense to unify main method and normal method for that case.

joshuabach: see my comment at line 141 for my solution to that

compiler-cli/integration-tests/semantic/single_static_method_not_main.mj.stderr
```diff
@@ -0,0 +1,9 @@
+error: Static methods have to be called 'main'
+   | 
+ 2 |     public static void s1(String[] args) {
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+ 4 |
```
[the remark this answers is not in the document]
reiner-dolp: Should go away with pull request #91

PR #87, opened 2018-11-13
Cli Tool: inspect-ast

reiner-dolp:


reiner-dolp:
Already reveals a lot of bugs.

Next steps would be
- [x] to use it for regression tests of spans
- [ ] print info from semantic analysis below code snippets
- [ ] use it for regression testing of semantic analysis phase
- [ ] implement the line/column command-line flag Christian and Joshua want
- [ ] fix the unbelievable amount of bugs uncovered in the parser and diagnostics printer :/

flip1995:
What are the found bugs?

reiner-dolp:
Incorrect line numbers in empty lines. Fixed in this pr. And if you run it on anything you will see that most spans are too small/wrong.

flip1995:
I fixed the postfix_op span bug. I couldn't find out why the span of classes starts on the second line and not on the `class` token, that is really weird.

reiner-dolp:
Bors r+

compiler-lib/src/asciifile/span.rs
```diff
@@ -274,6 +274,36 @@ impl<'span, 'file> Iterator for LineIterator<'span, 'file> {
     }
 }
 
+/// Iterator adapter that adds line numbers to a `LineIterator`
+///
+/// ```
+/// use compiler_lib::asciifile::{AsciiFile, Position, Span};
```
problame: does `cargo test` test that rustdoc examples compile?

compiler-lib/src/asciifile/span.rs
```diff
@@ -274,6 +274,36 @@ impl<'span, 'file> Iterator for LineIterator<'span, 'file> {
...
+///     vec![
+///         (1, "abcd\n".to_string()),
+///         (2, "\n".to_string()),
+///         (3, "\n".to_string()),
+///         (4, "efgh\n".to_string()),
+///         (5, "ijkl".to_string()),
+///     ]
+/// );
+/// ```
 pub struct LineNumberIterator<'span, 'file> {
```
problame: could we use `LineNumberIterator` [here](https://github.com/comprakt/comprakt/blob/fb51da6e2ba8ce576378a0ffbf6376648561c8e7/compiler-lib/src/asciifile/position.rs#L428) and [here](https://github.com/comprakt/comprakt/blob/fb51da6e2ba8ce576378a0ffbf6376648561c8e7/compiler-lib/src/asciifile/position.rs#L459) ?

inspect-ast/README.md
```diff
@@ -0,0 +1,61 @@
+Programmatically or interactively inspect the AST nodes before lowering to
+libFirm.
+
+Invoke with `cargo run --bin inspect-ast` or `cargo run -p inspect-ast`.
+
+
+```
```
problame: A reference to the existence of `--help` is sufficient IMHO.

inspect-ast/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+fn main() {
+    let cmd = CliCommand::from_args();
+
+    if let Err(msg) = do_main(cmd) {
+        exit_with_error(&msg);
+    }
+}
+
+fn do_main(cmd: CliCommand) -> Result<(), Error> {
```
problame: refactor this as soon as the *compiler controller* work is done https://comprakt.slack.com/archives/CDEADFB6F/p1542194320005700

problame:
>  I couldn't find out why the span of classes start on the second line and not on the class token. That is really weird.
In particular, it's only the first class delcaration. Try a file with multiple ones, the second one is spanned correctly. cc @flip1995 @reiner-dolp

flip1995:
>  the second one is spanned correctly

Is a newline before the class_decl sufficient?

problame:
Yes, it is, probably related to https://github.com/comprakt/comprakt/issues/88

compiler-lib/src/asciifile/span.rs
reiner-dolp: yes.

compiler-lib/src/asciifile/span.rs
reiner-dolp: We could :)

inspect-ast/src/main.rs
reiner-dolp: Yeah. The controller/driver is harder to do than expected :)

reiner-dolp:
Ehm, so whats a test case to reproduce this?

PR #89, opened 2018-11-14
add 'nested empty else' test to pretty-printer

reiner-dolp:


compiler-cli/integration-tests/ast/nested_empty_else.mj.stdout
```diff
@@ -0,0 +1,8 @@
+class A {
+	public void f() {
+		if (a)
+			if (b) { }
+			else ;
```
flip1995: >  Falls der else-Teil eines if-Statements ein leeres Statement enthält, so wird der komplette else-Teil
nicht ausgegeben.

```suggestion
```

compiler-cli/integration-tests/ast/nested_empty_else.mj.stdout
```diff
@@ -0,0 +1,8 @@
...
+		else i = 42;
```
flip1995:
```suggestion
		else 
			i = 42;
```

flip1995:
This is a specification problem, not a problem with our PP

reiner-dolp:
It's a bug since the output has different semantics.

compiler-cli/integration-tests/ast/nested_empty_else.mj.stdout
reiner-dolp: That's exatcly what the test is trying to assert. It should NOT be removed.

flip1995:
Hm. The fix would be to pretty print an if statement differently when the outer if statement has an else case.

reiner-dolp:
Do we have test that ensures it still works for not nested if-else blocks? Otherwise LGTM. Can be merged.

PR #91, opened 2018-11-14
Implicit row and column storage

reiner-dolp:
- Only saves newlines and looks up row and column of positions in O(log N) using binary search.
Simplifies the code a lot (and I probably haven't found all the code sections that can be simplified after this patch yet.)
- This cuts the size of spans and positions in half. Since we have a lot of spans and positions, this might be beneficial.
- Changes our newline convention, which was the devil!


reiner-dolp:
@problame r+?

reiner-dolp:
bors r=problame

PR #94, opened 2018-11-15
Implementation of Type-Checking/Attribution

hediet:
This introduces some duplicate checks, e.g. usage of main parameters, that have to be deduplicated in a a later commit.


problame:
merged in bc9190c fallout fixed in 1b5ccbf

Issue #84, opened 2018-11-13
Tracking issue: What to take care of in the semantic analysis


flip1995:
If something is being worked on in a PR or issue, link to it in the respective enum item.

---

During Name Analysis:
- [x] Exactly one static function
  - [x] static function  is called `main`
  - [x] argument of static function is used in `main` body
  - [x] argument of static function is redefined in `main` body
  - [x] this pointer inside static methods
  - [x] access on non-static methods
  - [x] access on non-static fields
- [x] ~If a class definition of `class String` exists, make sure it is used in the type check of `main`'s arguments.~ (Main arg as `String` is a syntactical feature in MJ) ~Java fails at runtime (not compile time) with 'Main method not found' error [demo here](https://repl.it/repls/EllipticalDimwittedNetworks))~ (In the semantical exceptions for MJ it says that exactly a single main method must exist => compile-time)
- [x] `ExpressionStatements` that are neither method calls nor assignments must be rejected

Before/During Type Check:
- [x] `main` must not be called
- [x] `System.out.println()` Edge cases
  - If `System` is undefined, use stdlib-system
  - If `System` is defined as a local variables => shadow (this is the easy case)
  - If `System` is defined as a type (`class System {}`) => fail (this is the edge case)
- [x] return inside functions on every path
- [x] `Expr` Type check.
- [x] l-value check

Other things:
- [x] Not a statement errors in java. 
- [x] `null` type ?
  - [x] comparisons of references to `null` must be allowed
  - [x] comparisons of primitves to `null` must be forbidden
  - [x] comparisons of arrays to `null` must be allowed
- [x] `void[][]`
- [x] try parse `IntLit` to `i32`

UB to handle:
- [x] ~~Usage of undefined variables (Definite Assignment Rule)~~ see #107 
- [x] ~~`null` access and div by 0~~ see #108 and #109 
- [x] ~~index out of bounds array access~~ see #110

reiner-dolp:
especially, exactly 1 main function exists.

flip1995:
Everyone can edit the OP

reiner-dolp:
Probably should also disallow redefining String.

reiner-dolp:
assigning to this should be disallowed.

flip1995:
Everything handled or moved to another Issue

PR #98, opened 2018-11-16
Add check that non-void methods always return

joshuabach:
See #84


compiler-lib/src/semantics.rs
```diff
@@ -124,6 +127,47 @@ impl<'a, 'f, 'cx> ClassesAndMembersVisitor<'a, 'f, 'cx> {
                         }
                         self.visit_static_block(&NodeKind::from(block), &params[0].name);
                     }
+
+                    match &member.kind {
+                        ast::ClassMemberKind::Method(ty, _, block)
+                            if ty.basic != ast::BasicType::Void =>
+                        {
+                            fn always_returns<'t>(stmt: &Spanned<'t, ast::Stmt<'t>>) -> bool {
```
flip1995: Move this function out of the match arm

compiler-cli/integration-tests/semantic/if_else_always_returns.mj.stderr
```diff
@@ -0,0 +1,13 @@
+error: method might not return
+    | 
+ 11 |       public int bar() {
```
reiner-dolp: this should probably mark the method signature (and as soon as multispans are supported, the offending block /subblock. In the current state the method name might not be displayed at all (neither in the error message, nor in the reproduced source snippet. E.g. code formatted as follows:

```
public int bar()
{

}
```

compiler-lib/src/semantics.rs
```diff
@@ -124,6 +127,47 @@ impl<'a, 'f, 'cx> ClassesAndMembersVisitor<'a, 'f, 'cx> {
...
+
+                                    // All other stmts do not always return
+                                    _ => false,
+                                }
+                            }
+
+                            if !block.statements.iter().any(always_returns) {
+                                self.context.diagnostics.error(&Spanned {
+                                    span: block.span.clone(),
+                                    data: SemanticError::MightNotReturn,
```
reiner-dolp: Add the method name to the error. It is available as `member.name` in the sourrounding scope

compiler-lib/src/semantics.rs
```diff
@@ -124,6 +127,47 @@ impl<'a, 'f, 'cx> ClassesAndMembersVisitor<'a, 'f, 'cx> {
...
+                                    ast::Stmt::Return(_) => true,
+
+                                    // All other stmts do not always return
+                                    _ => false,
+                                }
+                            }
+
+                            if !block.statements.iter().any(always_returns) {
+                                self.context.diagnostics.error(&Spanned {
+                                    span: block.span.clone(),
```
reiner-dolp: The span for the method declaration does not seam to be directly available. But it can probably be reconstructed out of `Method(
        Spanned<'t, Type<'t>>,
        Spanned<'t, ParameterList<'t>>,
        Spanned<'t, Block<'t>>,
),` by combining the spans of Type and ParameterList?

reiner-dolp: (This specific suggestion is not blocking.)

reiner-dolp: If you want to keep the block in the span, at least extend it to the signature by combining the first member of method (return type) with the block.span

compiler-cli/integration-tests/semantic/if_else_always_returns.mj
```diff
@@ -0,0 +1,18 @@
...
+         return 42;
+      }
+
+      public int bar() {
+         while (true) {
+           return 5;
+         }
+      }
+
+      public static void main(String[] args) {}
```
reiner-dolp: Btw. Shouldn't there be a test where the void returning method is not the special case "main method"

flip1995: Hm? Every method can have the return type `void`

compiler-lib/src/semantics.rs
```diff
@@ -124,6 +127,47 @@ impl<'a, 'f, 'cx> ClassesAndMembersVisitor<'a, 'f, 'cx> {
...
                         self.visit_static_block(&NodeKind::from(block), &params[0].name);
                     }
+
+                    match &member.kind {
+                        ast::ClassMemberKind::Method(ty, _, block)
+                            if ty.basic != ast::BasicType::Void =>
+                        {
+                            fn always_returns<'t>(stmt: &Spanned<'t, ast::Stmt<'t>>) -> bool {
+                                match &stmt.data {
+                                    // An if-else stmt always return iff both arms always return
```
problame: s/return/returns

compiler-lib/src/semantics.rs
```diff
@@ -124,6 +127,47 @@ impl<'a, 'f, 'cx> ClassesAndMembersVisitor<'a, 'f, 'cx> {
...
+                                            .as_ref()
+                                            .map_or(true, |else_arm| always_returns(&*else_arm));
+
+                                        then_arm_always_returns && else_arm_always_returns
+                                    }
+
+                                    // A block always returns iff any of the statements always
+                                    // returns
+                                    ast::Stmt::Block(block) => {
+                                        block.statements.iter().any(always_returns)
```
problame: i have found a test case where this is wrong, will push a fix

problame: see 865e858b71261e972bfb379ce4341c82b7efbe06

compiler-cli/integration-tests/semantic/if_else_always_returns.mj.stderr
problame: @reiner-dolp should we just add a TODO in the code for now? where is the multispan work being done / reviewed?

reiner-dolp: right now on my PC ;) i would just default to spanning the signature

compiler-lib/src/semantics.rs
problame: https://github.com/comprakt/comprakt/pull/98/commits/5a827df2904630928972852920d75cd368462eaf adds this, but it's pretty ugly since we currently do not have a link from `ClassMember` to its span.
The alterantive to the HashMap proposed in above commit to add a copy of a `ClassMember`'s full span (`ClassDecl.members[i].span`)  to the ClassMember.
Or we could change the visitor to also pass spans for each child, e.g. as a second closure argument. However, I think we only really _need_ this for `ClassMember`, both in first and second pass (second pass for mismatched return types, etc).

compiler-cli/integration-tests/semantic/if_else_always_returns.mj.stderr
problame: ok, ffr: spanning the signature is what we do since 5a827df2904630928972852920d75cd368462eaf

PR #99, opened 2018-11-16
symtab: add tests, docs, disallow shadowing, add benchmark for lookup

problame:
amortized O(1) lookup using hashmap is in a separate branch, but doesn't look promising in the benchmark: https://github.com/comprakt/comprakt/commit/a241717533fc46f31607c36a6d4156c3538c846f

@hediet current tip of `hedit/type-checking` doesn't compile, hence I can't rebase this

### DIFF
--- a/compiler-lib/benches/symtab_bench.rs
+++ b/compiler-lib/benches/symtab_bench.rs
@@ -1,0 +1,27 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+use compiler_lib::symtab::Scoped;
+
+fn scope_walk(input: &'_ usize) {
+    let defs: Vec<String> = (0..*input).map(|i| format!("scope#{}", i)).collect();
+    let mut scoped = Scoped::new();
+    for def in defs.iter() {
+        scoped.define(def, ()).unwrap();
+        scoped.enter_scope();
+    }
+    for depth in (0..*input).rev() {
+        let def = &defs[depth];
+        scoped.visible_definition(def);
+    }
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function_over_inputs(
+        "scope_walk",
+        |bencher, input| bencher.iter(|| scope_walk(input)),
+        vec![1, 10, 50, 100, 500],
+    );
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
--- a/compiler-lib/src/symtab.rs
+++ b/compiler-lib/src/symtab.rs
@@ -1,11 +1,17 @@
-use crate::strtab::Symbol;
 use std::collections::HashMap;
 
-pub type SymbolTable<'a, T> = HashMap<Symbol<'a>, T>;
+/// SymbolTable associates a Symbol `S` with a stored value `T`.
+pub type SymbolTable<S, T> = HashMap<S, T>;
 
-pub struct Scoped<'a, T> {
-    root: SymbolTable<'a, T>,
-    scopes: Vec<SymbolTable<'a, T>>,
+/// Scoped implements scoping for SymbolTable.
+/// The generic type `S` is the Symbol and `T` is the value stored for that
+/// symbol.
+pub struct Scoped<S, T>
+where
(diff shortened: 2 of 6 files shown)


compiler-lib/benches/symtab_bench.rs
```diff
@@ -0,0 +1,27 @@
...
+
+fn scope_walk(input: &'_ usize) {
+    let defs: Vec<String> = (0..*input).map(|i| format!("scope#{}", i)).collect();
+    let mut scoped = Scoped::new();
+    for def in defs.iter() {
+        scoped.define(def, ()).unwrap();
+        scoped.enter_scope();
+    }
+    for depth in (0..*input).rev() {
+        let def = &defs[depth];
```
problame: we should lookup `defs[0]`, not `defs[depth]`. this only the subsequent `.visible_definition()` should also be checked. I think it's returning None except for the first iteration of the loop.

compiler-lib/src/symtab.rs
```diff
@@ -32,18 +41,20 @@ impl<'a, T> Scoped<'a, T> {
...
+        debug_assert!(
+            insert_res.is_none(),
+            "visible_defintions broken, current_scope contained a value"
+        );
         Ok(())
     }
 
-    pub fn visible_definition(&self, sym: Symbol<'a>) -> Option<&T> {
-        // TODO O(1) access to visible definition
+    pub fn visible_definition(&self, sym: S) -> Option<&T> {
```
hediet: Btw, why is this method called like that? I'd prefer `get_visible_definition` or simpler `lookup`.

problame: it is on TIP of this branch, which didn't compile

compiler-lib/benches/symtab_bench.rs
problame: I'll improve this microbenchmark a little, and add another one that works on an actual AST.

problame: With `defs[0]`and the construction of `scoped` outside of the benchmark, we get positive time per op:

```
cope_walk/1            time:   [58.219 ns 58.449 ns 58.705 ns]                         
                        change: [-7.4808% -6.8786% -6.3780%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
scope_walk/10           time:   [594.27 ns 602.37 ns 612.59 ns]                           
                        change: [-81.014% -80.831% -80.613%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
scope_walk/50           time:   [2.9521 us 2.9720 us 2.9941 us]                           
                        change: [-95.784% -95.751% -95.719%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
scope_walk/100          time:   [5.9262 us 5.9879 us 6.0551 us]                            
                        change: [-98.002% -97.952% -97.905%] (p = 0.00 < 0.05)
                        Performance has improved.
scope_walk/500          time:   [29.092 us 29.215 us 29.359 us]                            
                        change: [-99.590% -99.586% -99.582%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

problame: OK, won't write actual AST-climbing microbenchmark, too much work.
But given above results,  I'll add the HashMap based lookups onto the target branch.

problame:
cherry-picked in 3d14ee90bd318344f99764af3011e3ed0bd626eb

compiler-lib/src/symtab.rs
reiner-dolp: rust convention is *not* using the "get_" prefix for getters and a "set_" prefix for setters.

PR #104, opened 2018-11-18
Fix panic with double classes

flip1995:
r? @hediet @joshuabach 

Easiest fix: only check the first definition of a class and completely ignore the following class definition inclusive members.


hediet:
Thanks :)
However, right now I am improving the type system so that the second class can also be checked without the need of merging class or ignoring types.

flip1995:
Yes it should be, since the second class is never checked. I'll add a test for it.

> right now I am improving the type system so that the second class can also be checked

Nice! I opened this PR with a quick fix, so we can merge it if you're not able to finish it until the deadline today.

hediet:
As of me, we can merge the quick fix right now. I can then rebase my changes.

flip1995:
No we cannot, it still panics on you're example. I'll look into this.

hediet:
@flip1995 It panics here:
```
                    let current_method = current_class
                        .get_method(member.name)
                        .expect("a class only has a member if it exists");
```
As a hotfix, just skip the loop if there is no method.

flip1995:
Oh no it doesn't panic. I tested it with the master branch, not with this branch, oops. :smile: 

I'll push the test.

flip1995:
I think we can merge this or close it, depending on the ETA of your fix @hediet.

hediet:
Give me an hour :)

flip1995:
bors r=hediet

PR #121, opened 2018-11-21
add `--lower` cli command for firm/assembler phase

reiner-dolp:
- integration tests will use this command for interfacing until the required command is released in next weeks assignment

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -72,21 +72,54 @@ enum CliCommand {
         #[structopt(name = "FILE", parse(from_os_str))]
         path: PathBuf,
     },
+    /// Output the AST as normalized MiniJava.
     #[structopt(name = "--print-ast")]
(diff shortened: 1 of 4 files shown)


joshuabach:
Technically, `--assemble` is the wrong name for this option. The process of "assembling" is converting assemblercode to machinecode (That's why the program doing it is called the *(GNU) assembler*). :trollface: I don't know if there's a good name for just producing assemblercode :-)

flip1995:
`--lower`

joshuabach:
> Oh we need to link the libfirm.so lib

If I'm not mistaken, waiting for problam/play-libfirm to be in a mergable state and then rebasing on that should suffice to resolve all build problems, right? At least `./run` works over there (doesn't work here)

flip1995:
cherry-picked in ontop of this. I want to move forward with this.

bors r+

PR #122, opened 2018-11-21
rename getter functions

flip1995:
See [API-guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html?highlight=getter#getter-names-follow-rust-convention-c-getter)

Soon this will be a Clippy lint: rust-lang-nursery/rust-clippy#1673


flip1995:
Let's queue this

bors r=reiner-dolp

PR #124, opened 2018-11-21
Run the test in the submission system tests in release mode

flip1995:
This avoids compiling twice, once in release and once in debug mode.


reiner-dolp:
good catch.

PR #125, opened 2018-11-21
build and deploy rust docs

reiner-dolp:


.travis.yml
```diff
@@ -93,3 +91,23 @@ script:
         return 1 ;
         ;;
     esac
+
+
+after_success: |
+    case "$TEST_KIND" in
+      comprakt-release)
+        cargo doc \
```
flip1995: `--document-private-items`?

Otherwise r=me

reiner-dolp: Yeah, I don't know, do we want that?

flip1995: otherwise documentation of every bin crate will be empty

reiner-dolp:
bors r=flip1995

.travis.yml
flip1995: image

vs only documentation for `LoweringOptions`

PR #130, opened 2018-11-22
Rusty abstractions for libfirm bindings by problame

flip1995:
Pretty much the same as `problame/play-libfirm`, but without the `play-libfirm` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,6 +418,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "libfirm-rs"
 version = "0.0.1"
 dependencies = [
+ "derive_more 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


problame:
I'm probably the worst candidate to review this because I wrote it:

* the `new_.*` methods on `Block` and `Graph` are incomplete -> we'll fix them as we go
  * the `new_(Add|Sub|...$ARITH_OP)` might be a pain point.
     * See issue https://github.com/comprakt/comprakt/issues/131
* the Node type-safety abstractions (all the wrapper structs around `*mut ir_node`) are probably not as well though-ought as they should be
  * this means users of `libfirm-rs` should not just try to work around Rust type-system compile errors when using `libfirm-rs`, but maybe question whether the Node type-safety abstraction is right

That said, let's merge this so the actual ir codegen work can base on this API.

Issue #117, opened 2018-11-19
Tracking Issue: libFirm IR generation


flip1995:
libFirm: [Website](http://pp.ipd.kit.edu/firm/), [GitHub](https://github.com/libfirm/libfirm), ([jFirm](https://github.com/libfirm/jFirm))

Everyone should write a libFirm program like in todays meeting with the generated bindings:
```java
class Foo {
    public static void main(String[] args) { }

    public int running_sum(int[] values, int[] output, int length) {
        int total = 0;
        int i = 0;
        while (i < length) {
            total = total + values[i];
            output[i] = total;
            i = i + 1;
        }
        return total;
    }
}
```
---

What to handle for the IR generation:

- [x] CLI integration #121 

- [x] libFirm bindings #28 

- [x] ~~Basic blocks generation~~ This shouldn't be necessary if we do the implementation of the lowering right

- [x] libFirm types
  - [x] Mapping AST/TypeSystem -> libFirm #141 
  - [x] Handle `boolean` and `void` (`boolean`: `u8`, `void`: only return -> no return arguments) 

Recommended steps by the task sheet:

- [x] libFirm function generation #134
  - [x] initial Firm-graph with just method nodes: all methods global (?) #134
  - [x] handle function params #134
  - [x] handle local vars #136
  - [x] handle `return` calls (depends on expr handling) #136 

- [x] libFirm expression handling (`boolean` handling?) #136
  - [x] map literal -> Firm nodes
  - [x] arithmetic expressions -> Firm graph
  - [x] local var and fn param access
  - [x] handle assignments
  - [x] address handling of field and array access + `load` & `store` node generation
  - [x] `call` expressions
  - [x] `new` expressions

- [x] Runtime system
  - [x] println, write, flush, read
  - [x] exception handling

- [x] x86-64 backend
  - lowering phase: transform `Sel` nodes -> address calculation
  - C runtime system lib to realize the runtime functions
  - Codegen with Firms x86-64 backend

joshuabach:
What's with program entry? Currently (e.g. in problam/play-libfirm), libFirm only produces an object file that can be linked to an object containing a main function. Or is this part of the next assignment?

flip1995:
Bugs detected by `mjtest-test`s test cases:
- [ ] SIGIOT: No such device or address [Travis](https://travis-ci.org/comprakt/comprakt/jobs/462176825#L1978) and 16 more
[...]

PR #134, opened 2018-11-22
 Implement libFirm function generation

flip1995:
cc #117 

When running
```
cargo run -- --lower Foo.java
```
Foo.java:
```java
class Foo {
    public static void main(String[] args) { }
}
```

This will generate
image

The `LoweringOptions` aren't used yet. Also it will segfault on every return type except of void.


problame:
bors try

PR #136, opened 2018-11-23
[WIP] libFirm method body code generation

joshuabach:
cc #117 

Closes #123 

~Based on #134, that should be merged before this.~

This can allegedly generate all control-flow structures ~(execpt `while`)~. The if-diamond seems to be generated correctly, variable declaration (of ints) and initial assignments also look fine. 

~Expression handling is very limited. Currently only supported are integer literals and local variable / parameter references. This is currently blocked on getting type information from @hediet.~

# TODO
- [x] `this` parameter
- [x] `while`
- [x] Expressions

PR #141, opened 2018-11-24
Hediet/type checking

hediet:
Some refactorings. Exposed Type Analysis results.


compiler-lib/src/semantics.rs
```diff
@@ -158,8 +158,8 @@ pub fn check<'a, 'f>(
     if context.diagnostics.errored() {
         return Err(());
     }
-
-    crate::type_checking::check(strtab, &ast, &context);
+    let mut type_system = crate::type_checking::type_system::TypeSystem::default();
```
flip1995: NIT: `use crate::type_checking::{self, type_system::TypeSystem};`

compiler-lib/src/type_checking/builtin_types.rs
```diff
@@ -0,0 +1,95 @@
...
+                    ty: writer_class_id.into(),
+                    can_write: false,
+                })
+                .unwrap();
+            type_system.add_class_def(system_class_def).unwrap()
+        };
+
+        context
+            .global_vars
+            .insert(strtab.intern("System"), system_class_id.into());
```
flip1995: Not `$System`? Or is `class System` not allowed. I'm still confused by this `String`/`System` inconsistency.

reiner-dolp:
CI fails because of several clippy warnings/errors.

compiler-lib/src/type_checking/builtin_types.rs
joshuabach: Nope. this is the name of the "global variable" `System`, not the *type/class* `$System`

joshuabach:
I'll tmp-merge this into my josh/firm-method-bodies to try it out

compiler-lib/src/type_checking/builtin_types.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl<'src> BuiltinTypes<'src> {
+    pub fn add_to<'ts>(
+        type_system: &'ts mut TypeSystem<'src>,
+        strtab: &'_ mut StringTable<'src>,
+        context: &'_ mut SemanticContext<'_, 'src>,
+    ) -> BuiltinTypes<'src> {
+        let int_ty = CheckedType::Int;
+
+        let (_, reader_class_id) = {
```
reiner-dolp: These `let (_, reader_class_id) = { /* function body */ };` are really weird. You should probably add a comment explaining why its written that way.

reiner-dolp: Could you add a few markdown paragraphs to the type_checking/mod.rs that explain implementation details like the special meaning of "$"?

problame: will take care of that

compiler-lib/src/type_checking/type_system.rs
```diff
@@ -168,7 +183,8 @@ pub enum CheckedType<'src> {
     Boolean,
     Void,
     Null,
-    TypeRef(Symbol<'src>),
+    TypeRef(ClassDefId<'src>),
+    UnknownType(Symbol<'src>),
```
problame: pleas add a comment *right here*  what this is, and whether we need to take care of it in the ir codegen.
if not, please change the public API such that this implementation detail is not leaked.

otherwise, to get exhaustive enum matches, we need to panic wherever we match on `CheckedType`

comments?

compiler-lib/src/type_checking/mod.rs
```diff
@@ -1,409 +1,32 @@
...
+//!
+//! MiniJava suffers from the fact that its specification was stripped down
+//! from the Java language report.
+//! A good example is the handling of the main method's arguments:
+//! `String[] args` must not be used, and according to the spec, `String[]`
+//! within the main method's parameter list is not a type.
+//! However, it is possible for users to define a `class String` in MiniJava,
+//! which ,in Java, would shadow the `java.lang.String` that is part of the
+//! main method's parameter type. Not so in MiniJava - at least that's the
+//! result of the discussion we had in the
```
reiner-dolp: NIT: @problame I think you have the same problem as me. Your default editor config brakes lines shorter than our rust format config, which results in these weird line breaks within paragraphs :)

compiler-lib/src/type_checking/mod.rs
```diff
@@ -1,409 +1,32 @@
...
+//! lecture.
+//!
+//! To accomodate this and other spec quirks, we define built-in types using
+//! `$String` or similar: Since a class definition beginning with `$` is
+//! rejected by the lexer, we can safely introduce such definitions during type
+//! analysis for the built-in
+//! types.
```
reiner-dolp: Same here :)

compiler-lib/src/type_checking/mod.rs
reiner-dolp: Thanks for the docs

compiler-lib/src/type_checking/type_system.rs
```diff
@@ -48,21 +49,39 @@ impl<'src> TypeSystem<'src> {
         self.defined_classes.get_mut(&name)
     }
 
-    pub fn lookup_class(&self, name: Symbol<'src>) -> Option<&ClassDef<'src>> {
-        self.defined_classes.get(&name)
+    pub fn lookup_class(&self, name: Symbol<'src>) -> Option<(&ClassDef<'src>, ClassDefId<'src>)> {
```
problame: In the ir codegen, we invoke this like the following:

```
let class_type = self.type_system.lookup_class(class.data.name.data);
```

Would be nicer to just invoke it with `class.data`, where `class` has type 

```
 `&asciifile::spanned::Spanned<'_, ast::ClassDeclaration<'_>>`                                                                                                                                                         
```

problame:
I experimented with the type checking API as described in https://github.com/comprakt/comprakt/pull/136#issuecomment-441396383

compiler-lib/src/type_checking/type_system.rs
flip1995: `UnknownType` is only constructed in `checked_type_from_ty` if `lookup_class` fails. This shouldn't happen at all IIUC.

compiler-lib/src/type_checking/type_system.rs
flip1995: This is not possible, since we also need to be able to call this function with `ty.name` and other symbols.

```bash
$ rg lookup_class
compiler-lib/src/type_checking/method_body_type_checker.rs
333:            NewObject(name) => match self.type_system.lookup_class(name.data) {
568:            match self.type_system.lookup_class(var_name.data) {

compiler-lib/src/type_checking/checker.rs
82:        let name = match type_system.lookup_class_mut(class_decl.name.data) {
233:        Custom(name) => match type_system.lookup_class(*name) {
```

flip1995:
Let's merge this and figure out how to get rid of the `UnknownType` in another PR.

bors r+

compiler-lib/src/type_checking/type_system.rs
joshuabach: See #143

compiler-lib/src/type_checking/type_system.rs
hediet: This is what the `TypeAnalysis` struct is for:
`type_analysis.decl_get_class_id(class_decl: &'ast ast::ClassDeclaration<'src>) -> Option<ClassDefId<'src>>`.

PR #158, opened 2018-12-03
Add test checking if while true timeouts

reiner-dolp:


reiner-dolp:
I am biggypacking an unrelated *.layout file ignore onto this.

PR #164, opened 2018-12-04
Cli Interface and Integration Tests for Optimization Phase

reiner-dolp:
- [x] Adds the command-line flag "--optimizations/-O" that can be used to specify a list of optimizations to apply. Currently defaults to no optimizations. This allows you to call the compiler as follows:
   ```
   cargo run -- --compile-firm input.mj  -O ConstantFolding,AlgebraicSimplification,ConstantFolding
   ```
   which will apply the optimization ConsantFolding, followed by an AlgebraicSimplification optimization, followed by another ConstantFolding pass.
- [x] Add a `integration-tests/optimization` folder that accepts mini java files with yaml front matter. For example:
    ```java
    ---
   optimizations: [AlgebraicSimplification]
   expect: change
   stdout:
      is: "4\n"
   ---
   class SimplifyMul1 {
       int x = 4;
       System.out.println(x * 1);
     }
   }
    ```
    will invoke the compiler with the `AlgebraicSimplification` optimization and assert the following:
    * the stderr/stdout/stdin of the generated binary did not change.
    * the generated assembly changed.
- [x] Since you cannot call the compiler directly using any of the files with yaml front matter. This PR provides a simple command-line tool `cargo run --bin strip-front-matter FILE` that removes the yaml front matter from a mini java file.
- [x] All tests can accept expected output and input in a separate yaml file or using a yaml front matter block. For example:
    ```java
    ---
    stderr:
      is: ''
    stdout:
      is: 'abcd'
    stdin:
      is: 'abcd'
    exitcode:
      is: 0
    ---
    class Cat {
            int c;
            while((c = System.in.read()) != -1) {
                System.out.write(c);
            }
        }
    }
    ```
   Lookup rules are as follows:
   1. Check if a yaml front matter block exists.
   2. Check if a separate file `TEST_NAME.yml` exists.
   3. Expect the output to be in separate files: `TEST_NAME.stderr` `TEST_NAME.stdout` `TEST_NAME.exitcode`

   if you want to place a part of inputs and outputs in the front matter and another part into separate files, you can use `file: PATH` instead of `is: STRING` to reference an external file.
- [x] Set the following defaults for test folders that test behaviour of the binary: `compiler_stdout = compiler_stderr = ""`, `compiler_exitcode = 0`.


reiner-dolp:
A few more examples to demonstrate the yaml syntax:
```java
---
optimizations:
    - ConstantFolding
    - AlgebraicSimplification
    - ConstantFolding
expect: Unchanged
---
class SimplifyMul1 {
}
```
```java
---
optimizations: [ConstantFolding,AlgebraicSimplification,ConstantFolding]
expect: Unchanged
---
class SimplifyMul1 {
}
```

hediet:
Looks very cool. However, what do you think of supporting yaml files directly besides mj files? You could use the file extension to tell them apart. IMHO we could also move some of the other test files into this new format so that we can specify error code etc. in a single file

reiner-dolp:
Needs cleanup, merging since basic functionality is given and needed.

bors r=joshuabach,hediet,reiner-dolp

PR #165, opened 2018-12-04
New Libfirm API, Basic Optimizations (Constant Folding, Unreachable Code Elimination)

hediet:
- Implements a new libfirm abstraction
- does however not update old code using the old libfirm abstraction
- implements optimization phase
  - constant folding for ints
  - unreachable code eliminiation, which effectively is constant folding for bools
- uses fixpoint iteration alternating between both optimizations


libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,2850 @@
...
+    Start(StartNode),
+    Store(StoreNode),
+    Sub(SubNode),
+    Switch(SwitchNode),
+    Sync(SyncNode),
+    Tuple(TupleNode),
+    Unknown(UnknownNode),
+}
+
+impl Into<*const ir_node> for Node {
```
flip1995: `From` instead of `Into`: 
> Library authors should not directly implement this trait, but should prefer implementing the From trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

https://doc.rust-lang.org/std/convert/trait.Into.html

libfirm-rs/generator/gen.ts
```diff
@@ -0,0 +1,269 @@
+import fs from "fs";
```
flip1995: Why typescript? :sob: This could be nicely done with serde-json and `proc_macro`s...

libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,2786 @@
+use libfirm_rs_bindings as bindings;
+use std::collections::HashMap;
+use super::nodes::NodeTrait;
+
+#[derive(Debug)]
+pub enum Node {
```
joshuabach: In the current `libfirm_rs` wrappers, each Node type (at least the ones we use) is a seperate struct, not a variant of one big `Node` enum. That makes it possible to define traits for "groups" of nodes, such as the `ValueNode` or `ALUOperand` traits, so that certain functions can only accepts Nodes implementing that trait. 

The advantage is that the compiler complains if you use the wrong node in the wrong place.

I suspect that this is more difficult (impossible) to completely autogenerate, because the information of which function accepts which kind of node is only menthioned in the libfirm docs (if at all)

libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,2786 @@
...
+            map.insert(op_code, Self::create_unknown);
+        }
+        NodeFactory(map)
+    }
+
+    pub fn node(ir_node: *mut bindings::ir_node) -> Node {
+        Self::new().create(ir_node)
+    }
+
+    pub fn create(&self, ir_node: *mut bindings::ir_node) -> Node {
```
joshuabach: Rust convention for constructor functions names is `new`

libfirm-rs/src/nodes.rs
```diff
@@ -0,0 +1,24 @@
...
+            Node::Block(block) => block,
+            _ => panic!("Expected block.")
+        }
+    }
+
+    // TODO implement methods from
+    // https://github.com/libfirm/jFirm/blob/master/src/firm/nodes/Node.java
+}
+
+// TODO maybe Into<*const ir_node> for NodeTrait?
```
joshuabach: Better not. If we are doing a full-blown abstraction now, don't leak internals

libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,2786 @@
...
+        let op_code = unsafe { bindings::get_irn_opcode(ir_node) };
+        let f = self.0.get(&op_code).unwrap();
+        f(ir_node)
+    }
+
+    fn create_add(ir_node: *mut bindings::ir_node) -> Node {
```
joshuabach: In the current libfirm_rs wrappers, each Node type (at least the ones we use) is a seperate struct, not a variant of one big Node enum. That makes it possible to define traits for "groups" of nodes, such as the ValueNode trait, so that certain functions can only accepts Nodes implementing that trait.

In this case, we would have sth like
```rust
impl NodeFactory {
   fn new_add<LHS, RHS>(lhs: LHS, rhs: RHS) -> Add
   where LHS: ValueNode, RHS: ValueNode { /*...*/}
   /*...*/
}

impl ValueNode for Add { /*...*/}
```

The advantage is that the compiler complains if you use the wrong node in the wrong place.

This of course also means that each of the `create_*` functions would need to return a specific node-struct and not an instance of `Node` enum.

I suspect that this is more difficult (impossible) to completely autogenerate, because the information of which function accepts which kind of node is only menthioned in the libfirm docs (if at all)

libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,2786 @@
...
+
+impl Add {
+    /// Gets first operand.
+    pub fn left(&self) -> Node {
+        let unwrapped = unsafe { bindings::get_Add_left(self.0) };
+        NodeFactory::node(unwrapped)
+    }
+
+    /// Sets first operand.
+    pub fn set_left(&self, val: &'_ Node) {
```
joshuabach: Maybe an atomic 
```rust
pub fn set_left<F>(&self, F)
where F: FnOnce(Node) -> Node { /*...*/ }
```
is better, to reduce the danger of leaking pointers to about-to-be-replaced `left` arm

libfirm-rs/src/nodes_gen.rs
hediet: I considered that. There are also structs for each node type, but there must be a way of constructing nodes in a generic way. That is what NodeFactory is for.

hediet:
@joshuabach In the next hour or so I will push construction methods. They were a little bit more work because they involved more data types than just nodes.

libfirm-rs/src/nodes_gen.rs
hediet: I don't fully understand. What is the danger and how is it avoided by that?

libfirm-rs/src/nodes_gen.rs
```diff
@@ -0,0 +1,4048 @@
...
+                    1 => ProjKind::Store_XRegular(node),
+                    2 => ProjKind::Store_XExcept(node),
+                    _ => ProjKind::Other,
+               },
+            Node::Switch(node) => 
+               match proj.num() {
+                    0 => ProjKind::Switch_Default(node),
+                    _ => ProjKind::Other,
+               },
+            Node::Proj(proj, ProjKind::Start_TArgs(start)) => ProjKind::Start_TArgs_Arg(proj.num(), start, proj),
```
reiner-dolp: possible endless recursion

libfirm-rs/src/nodes_gen.rs
reiner-dolp: he is suggesting a replace function.

libfirm-rs/src/types.rs
```diff
@@ -0,0 +1,230 @@
...
+                match self {
+                    $($name::$var(x) => x.ir_type(),)*
+                }
+            }
+        }
+
+        $(
+            #[derive(Clone, Copy)]
+            pub struct $ty(*mut bindings::ir_type);
+            impl $ty {
```
reiner-dolp: TryFrom

libfirm-rs/src/types.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+impl ClassTy {
+    pub fn new_class_type(name: &str) -> ClassTy {
+        ClassTy::from(
+            Ty::from_ir_type(
+                unsafe {
+                    bindings::new_type_class(
+                        CString::new(name)
```
reiner-dolp: this is freed to early since you are only holding a pointer.

libfirm-rs/src/types.rs
```diff
@@ -0,0 +1,230 @@
...
+                Ty::Other(OtherTy(ty))
+            } else {
+                Ty::Other(OtherTy(ty))
+                // unknown type kind found
+            }
+        }
+    }
+}
+
+trait TyTrait : Sized {
```
reiner-dolp: Sized can be removed

reiner-dolp:
The generator should surpress some warnings by adding the prefix:

```
#![allow(dead_code)]
#![allow(clippy::all)]
```

And run rustfmt after generation.

reiner-dolp:
The generator should surpress some warnings by adding the prefix:

```
#![allow(dead_code)]
#![allow(clippy::all)]
```

And run rustfmt after generation.

reiner-dolp:
The generator should surpress some warnings by adding the prefix:

```
#![allow(dead_code)]
#![allow(clippy::all)]
```

And run rustfmt after generation.

reiner-dolp:
Instructions for other people on running the generator:
```
# get deps (do this once)
yarn global add ts-node 
cd libfirm-rs/generator
yarn 
# do this each time the generator script is changed
ts-node ./gen.ts
cargo fmt --all
```

libfirm-rs/src/nodes_gen.rs
reiner-dolp: per discussion with andreas and sebastian today, this is guranteed to be impossible by libfirm internals.

reiner-dolp:
Preserving @hediet old PR description:

---
This PR implements an improved libfirm api. Work in progress.

The rust code is generated from a json file which is extracted from libfirms' python script.

Todo:
* ~Generate types for all nodes~
* Generate typed Project Pins from `data.json` that allow to match on their parent.
* Generate Construction class from Java api. 
* Integrate with `lib.rs` datastructures (`Graph`, `Entity`, `Types`, ...).
* Update existing code to use new types

reiner-dolp:
bors r+

There is general consent to merge this, even though there is additional work to be done:

- nicer cli interface
- no fixpoint iteration +  abstraction for optimizations

Also see tracking issue #171.

PR #166, opened 2018-12-04
Add test for eval order and side effects examples + bugfix

flip1995:
With the examples of the test our compiler is consistent with javac. I will continue to search for more exec order and side effect tests.

Currently only the tests from the slides are implemented.

If anyone has an idea for a test case feel free to add it.


flip1995:
I gone through the JLS "Evaluation Order" chapter and copied/generated tests from there. Didn't get to verify them yet. Will do tomorrow.

integration-tests/binary/jls_15_13_1_2.out.exitcode
```diff
@@ -0,0 +1 @@
+terminated by signal (or crashed)
```
problame: doesn't seem to me like our test framework can recognize this (?)

reiner-dolp: What do you mean?

reiner-dolp: https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code

flip1995:
ETA Wednesday eve

Doesn't seem urgent to me. The optimizations shouldn't depend on this.

integration-tests/binary/jls_15_13_1_2.out.exitcode
problame: I thought only integers were allowed... ?

reiner-dolp: So, is this about the test case being wrong? or about the integration test framework having a bug.

reiner-dolp: The exitcode file is compared by string. So you can actually write to it what you want. In case the program terminates abnormally (and therefore does not return an exit code), the string shown above is printed by the test framework.

flip1995:
Well that was a quick and easy fix :smile: 

r? @joshuabach

PR #167, opened 2018-12-04
move integration tests to top level dir

reiner-dolp:


runner-integration-tests/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition = "2018"
+
+[dependencies]
+integration_test_codegen = { path = "../codegen-integration-tests" }
+difference = "2.0"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+lazy_static = "1.2.0"
+wait-timeout = "0.1.5"
```
flip1995: Do we need serde, lazy_static and wait-timeout in this crate?

reiner-dolp: wait-timeout is used for `timeout` tests. `lazy_static` is used for caching cargo calls.
I think serde is indeed unused until #164 is merged.

reiner-dolp:
bors r=flip1995

PR #173, opened 2018-12-13
Improved constant folding

hediet:
WIP PR for improved constant folding that has less bugs, is faster and does dead code elemination.


integration-tests/optimization/dead_code_elim_with_constant_folding.mj
```diff
@@ -0,0 +1,32 @@
+---
+optimizations: [ConstantFolding]
+expect: change
+compiler_optimized_stdout: ignore
```
reiner-dolp: you can remove these ignores if you do not have debug output using println!.

PR #182, opened 2018-12-20
Improvements to Debugger GUI

hediet:
- [x] generalize from Methods to arbitrary Graphs
- [x] filters runtime functions by default
- [x] adds ENV variables `FILTER_BREAKPOINT_LABEL=[label,]*`, `FILTER_BREAKPOINT_LOCATION=[file:line,]*`, `FILTER_GRAPH_NAME=[name,]*`
- [x] add REST endpoints  to add filters
- [ ] ~add GUI buttons to filter graphs and breakpoints~ Might be implemented in a future PR.

### DIFF
--- a/compiler-lib/src/debugging.rs
+++ b/compiler-lib/src/debugging.rs
@@ -45,21 +45,23 @@
 //! message content is the current compiler state.
 
 use crate::dot::GraphState;
-use rocket::{get, http::Status, State};
+use rocket::{
+    fairing::{Fairing, Info, Kind},
+    get,
+    http::{ContentType, Header, Method, Status},
+    Request, Response, State,
+};
 use rocket_contrib::{json::Json, serve::StaticFiles};
 use serde_derive::Serialize;
 use std::{
     collections::hash_map::HashMap,
+    io::Cursor,
     sync::{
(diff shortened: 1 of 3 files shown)


compiler-lib/src/debugging.rs
```diff
@@ -223,6 +226,31 @@ impl Debugger {
 
 struct DebuggerState(Debugger);
 
+pub struct CORS();
```
flip1995: NIT:
```suggestion
pub struct CORS;
```

compiler-lib/src/optimization/constant_folding.rs
```diff
@@ -147,6 +147,27 @@ impl ConstantFolding {
...
+            }
+
+            if node == cur_node {
+                label = label
+                    .style(Style::Filled)
+                    .fillcolor(X11Color::Blue)
+                    .fontcolor(X11Color::White);
+            }
+
+            if let Some(priority) = self.queue.get(&node) {
```
flip1995:
```suggestion
            if let Some(_) = self.queue.get(&node) {
```

hediet:
@reiner-dolp Fixed, even though I considered these issues irrelevant as it is somewhat hacky anyways.

PR #185, opened 2018-12-29
New libfirm generation

hediet:
Is based on #184. Reimplementation of libfirm IR generation using new api.
The high level structure is roughly the same as before.
However, `LastBlockJumps` is not needed anymore because `ActiveBlock` tracks whether there is an active block that still needs a jump target.


libfirm-rs/src/graph.rs
```diff
@@ -225,6 +227,53 @@ impl Graph {
             NodeFactory::node(new_node_ptr)
         }
     }
+
+    // == Construction ==
+    /*
+    pub fn value(self, slot_idx: usize, mode: Mode) -> Node {
```
flip1995: Note: Moved to `libfirm-rs/src/nodes.rs`

compiler-lib/src/firm/method_body_generator.rs
```diff
@@ -1,140 +1,137 @@
...
-
-impl LastBlockJumps {
-    fn jumps(self) -> bool {
-        use self::LastBlockJumps::*;
-        match self {
-            Yes => true,
-            No => false,
-        }
-    }
+enum ActiveBlock {
```
flip1995: Why? 
```rust
type ActiveBlock = Option<Block>
```
would do the trick, wouldn't it?

flip1995:
Superseded by #186

libfirm-rs/src/graph.rs
flip1995: Removed in #186

libfirm-rs/src/tarval.rs
```diff
@@ -92,8 +98,12 @@ impl Tarval {
     }
 
     #[inline]
-    pub fn mode(self) -> bindings::mode::Type {
-        unsafe { bindings::get_tarval_mode(self.0) }
+    pub fn mode(self) -> Mode {
+        Mode::from_libfirm(unsafe { bindings::get_tarval_mode(self.0) })
+    }
+
+    pub fn mode_ex(self) -> Mode {
```
flip1995: Same as `mode(self)` above

compiler-lib/src/firm/method_body_generator.rs
flip1995: Won't fix, discussed on slack

PR #189, opened 2019-01-01
various improvements

hediet:
- [x] make abstract graph convertible to dot/debugger format, instead of firm graphs only.
- [x] fix imports and remove uniform_paths
- [x] add maps, eq, hash on entity


libfirm-rs/src/entity.rs
```diff
@@ -62,3 +65,18 @@ impl Entity {
         }
     }
 }
+
+#[allow(clippy::derive_hash_xor_eq)]
```
reiner-dolp: this allow is not required? your implementing both by hand.

hediet: Fixed.

reiner-dolp:
part of #192

PR #191, opened 2019-01-01
Custom Log Format

hediet:


hediet:
> Could you move the logging closure out of main into its own function?

Fixed.

PR #192, opened 2019-01-01
Improved constant folding and other things

hediet:
Merge and review #190 first


reiner-dolp:
Bisect shows that the regression is introduced in: Implemented 0*x = 0 optimization. 6da9ac9

PR #193, opened 2019-01-02
[WIP] Fix Inlining

reiner-dolp:
- [x] make 'Inline,ConstantFolding' the default
- [x] fix inlining


compiler-lib/src/optimization/constant_folding.rs
```diff
@@ -59,6 +61,8 @@ impl ConstantFolding {
         let mut node_topo_idx = HashMap::new();
         let mut values = HashMap::new();
 
+        graph.assure_outs();
```
reiner-dolp: this is pretty expensive given that its only used for debugging. Should be in a `if cfg!("graphical_debugger") { }`

PR #198, opened 2019-01-10
[WIP] Register allocation, the ending

flip1995:

PR #206, opened 2019-01-17
cli: some code dedup

problame:
`firm::build` still requires refactoring, but that work will happen on #198  `phil/calling-convention`.


reiner-dolp:
This fails clippy checks

problame:
There's also a race with tempdir if the emit-asm flag is not set.

problame:
bors try

PR #208, opened 2019-01-18
cli/firm: refactor: compilation steps + multiple backends

problame:
FIRM: mod.rs refactoring:

- Goal: split firm::build up into steps such that we can have
  another backend than libFIRM that consumes the FirmProgram
- `firm::build` => `FirmContext::build`
- FirmContext is a singleton, which is the best we can do
  given libFIRM's internal global state (target config)
- FirmContext now only abstracts away the things we do on the FIRM
  graph, in the following order:

  - construction from AST
  - pre-optimization ("high-level") dumping
  - running optimizations on the graph
  - EITHER: use libFIRM as backend
      - we implement AsmBackend for that
  - OR: return FirmProgram, which can be used by an external backend
      - the lowering WIP branch will provide an AsmBackend as well

- Above order is enforced by a state machine (FirmContextState)
  using exhaustive matching at the beginning of each public function
  of FirmContext

CLI restructuring:

- Prepare `--compile` subcommand, and prepare it to support both
  an AMD64 backend and a Molki backend.

Reusable Binary Generation from Assembly:

- binary-production logic was extracted to BinaryGenerator
- A BinaryGenerator uses an AsmBackend to produce a binary
- An AsmBackend can emit assembly that is understood by `cc`

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -26,7 +26,10 @@
 use compiler_lib::{
     asciifile, ast,
     context::Context,
-    firm,
+    firm::{
+        self,
+        runtime::{self, RTLib},
+    },
     lexer::{Lexer, TokenKind},
     parser::Parser,
     print::{self, lextest},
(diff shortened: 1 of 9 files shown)


problame:
@flip1995 code for lowering that integrates the current backends is ready, will commit that onto lowering as soon as we merge this change into master, then lowering

compiler-cli/src/main.rs
```diff
@@ -114,60 +117,112 @@ pub enum CliCommand {
...
+    /// backend. Defaults to 'a.out' in the current working directory.
+    #[structopt(name = "--compile")]
+    Compile(CompileOptions),
+}
+
+/// Backends supported by the [`CliCommand::Compile`] (`--compile`) command.
+#[derive(StructOpt, Debug, Clone)]
+pub enum CompileBackend {
+    #[structopt(name = "amd64")]
+    AMD64,
```
flip1995: This is called
```suggestion
    Amd64,
```

in the lowering branch. This should be consistent

compiler-cli/src/main.rs
```diff
@@ -114,60 +117,112 @@ pub enum CliCommand {
...
+    pub dump_assembly: Option<PathBuf>,
+
+    /// Write binary to the given file or directory
+    #[structopt(long = "--output", short = "-o", parse(from_os_str))]
+    pub output: Option<PathBuf>,
+}
 fn main() {
     build_logger().init();
 
+    libfirm_rs::init();
```
flip1995: For reference

flip1995:
> that integrates the current backends is ready

could you push that on a separate branch (pointing to lowering)?

problame:
> > that integrates the current backends is ready
> 
> could you push that on a separate branch (pointing to lowering)?

see #209

compiler-lib/src/firm/mod.rs
```diff
@@ -58,108 +55,192 @@ pub enum FirmError {
...
+        if prev {
+            panic!(
+                "libFIRM lowering context initialized, concurrent or \
+                 repeated use not supported due to library-internal state"
             );
-        }
-    }
+        } else {
+            unsafe {
+                libfirm_rs::init();
```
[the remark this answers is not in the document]
problame: yes, but it's a Once, thus this second call has no effect but is semantically important if compiler-lib were ever used by another new CLI crate.

flip1995:
Nice, let's get this going

bors r+

PR #209, opened 2019-01-19
lowering: suuport new CLI

problame:


problame:
flag order for `--compile` is broken in 6fc798a, don't merge

PR #172, opened 2018-12-13
[WIP] Lowering

joshuabach:
Blocked on #165 

*Very* WIP

PR #210, opened 2019-01-21
Integration tests using own backend

reiner-dolp:
Run all integration tests with the libfirm backend and our own backend. (Previously only the libfirm backend was used.)


flip1995:
```java
---
stdin:
    file: cat.out.stdin
---
class Cat {
        int c;
        while((c = System.in.read()) != -1) {
            System.out.write(c);
        }
    }
}
```

Results in 
`test binary_Backend____Own::cat_mj ... test binary_Backend____Own::cat_mj has been running for over 60 seconds`

Same for example_yml.mj test. Is something wrong with the yml parsing or is our backend faulty?

reiner-dolp:
Since ` binary_Backend___Libfirm` works its most likely the backend

reiner-dolp:
will be merged indirectly in #172

PR #211, opened 2019-01-21
build/jenkins: initial support

problame:
Build times with Jenkins

* No cache (new branch / PR):  13min
* Cached, idle build node: 4 min
* Cached, busy build node: 7 min

I'm not _that_ confident about the quality of the cache (i.e. equivalence to clean build), but we can still gate `master` with bors + travis integration.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.cargo/
+target/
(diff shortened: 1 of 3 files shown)


flip1995 (approved):
Approve, so it disappears from my "review requested" list.

PR #213, opened 2019-01-21
First version of load-store optimization in constant folding.

hediet:


compiler-lib/src/optimization/constant_folding.rs
```diff
@@ -257,25 +369,39 @@ impl ConstantFolding {
                     },
                 )
             }
+
+            // == Value nodes ==
             _ => {
```
reiner-dolp: No need to change this. But you could reduce the nesting here by matching on `_ if condition =>`.

hediet: Ah I see. I feared that rust wouldn't do a jump table anymore with guarded matches.

compiler-lib/src/optimization/lattices/heap.rs
```diff
@@ -0,0 +1,371 @@
...
+        Self {
+            p: self.p,
+            recent: true,
+        }
+    }
+}
+
+impl fmt::Debug for Pointer {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "🆕{}{}", self.p, if self.recent { "r" } else { "" })
```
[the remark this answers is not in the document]
hediet: Yeah, for example mine :D But it looks very nice in the graph debugger :D If you want, I can change it to ascii character as I hopefully don't need to debug it that often anymore.

hediet:
> * Being able to toggle each optimization on its own would be pretty useful. Do you have an idea how this could be implemented?
Not really, in particular not how to run load-store-optimization without constant folding.
I could guard the section "== load store ==" to make the load-store-optimization configurable.

> * Please drop the non-ascii char from the debug output. Then bors r=reiner-dolp.
Done.

hediet:
bors r=reiner-dolp

PR #214, opened 2019-01-21
Implemented debugger helpers for debugging in vscode with visualizer extensions and gdb

hediet:


compiler-lib/src/debugging.rs
```diff
@@ -63,36 +63,108 @@ use std::{
     thread::{self, JoinHandle},
 };
 
-#[cfg(feature = "debugger_gui")]
+#[cfg(any(feature = "debugger_gui", feature = "debugger_vscode"))]
```
reiner-dolp: since both breakpoint implementations have nothing in common, why not make a seprate #[cfg(feature = "debugger_vscode")] implementation?

compiler-lib/src/debugging.rs
```diff
@@ -63,36 +63,108 @@ use std::{
...
+        // to get these methods generated so that they can be used for debugging
+        get_all_graph_data();
+        get_last_graph_data();
+    }
+    let mut history = LAST_PROG.lock().unwrap();
+    history.items.push(program);
+}
+
+/// use for debugging
+pub fn get_all_graph_data() -> String {
```
reiner-dolp: is this "pub" to surpress the unused warning? allow(dead_code) would be more explicit.

hediet: This is pub so I can call it from gdb. But I don't know whether this conclusion is true.

compiler-lib/src/debugging.rs
hediet: They have a lot in common. This way, you can use the debugger server and I can use the debugger visualizer in vs code without needing to change anything or having code duplicates.

reiner-dolp: What I mean is this:

```
#[cfg(feature = "debugger_gui")]
macro_rules! breakpoint {
// all the stuff from the if-truee-branches in the current impl
}
#[cfg(feature = "debugger_vscode")]
macro_rules! breakpoint {
// all the stuff from the else-branches in the current impl
}
```

compiler-lib/src/debugging.rs
reiner-dolp: ok. thanks

PR #215, opened 2019-01-23
Refactor compiler-lib (split into separate crates)

problame:
* Primary Goal: Minimize Build time while working on 'hot' files, i.e., code that is still changing rapidly
* Solution: Split off compiler-lib modules that are basically done
  * lexer
  * parser
  * type_checking
  * firm construction
* Move common structs / fns to utils / compiler_shared
* minimize number of changed lines by `use`ing dependencies at the crate level
  * can refactor that later

**Try this** / merge your current WIP branch and see what breaks!
During review, use GitHub's suggest edit feature.


reiner-dolp:
this needs a benchmark before it can be merged.

problame:
> this should be benchmarked before it's merged.

Two things this change affects:

1. build time for building compiler-cli
  * this is what we're aiming to reduce here
2. run time of compiler-cli due to less inter-proc-optimizations taking place inside rustc when compiling compiler-lib / compiler-cli (crates are the unit of translation, right?)

I would suggest you just test-merge this branch into your development branch, work as usual and check if compile times (aspect No. 1 above) improve.

PR #217, opened 2019-01-23
WIP: Code placement and Common Subexpression Elimination

reiner-dolp:
- Implements CodePlacement à la Libfirm with some changes
  - [x] use frequency analysis instead of loop nesting depth as placement heuristic
  - [x] since CodePlacement computes all valid placements for an expression, perform
    common subexpression elimination as an additional heuristic. Prefer placements
    that can eliminate expressions. This effectively results in global common subexpression elimination using a block-local common subexpression elimination
  - [ ] (optional) only perform GCSE in some cases: e.g. Subexpression A dominates Subexpression B, or Subexpression B dominates Subexpression A, maybe if both expressions post dominate a common block? This needs some testing.


reiner-dolp:
bors r=problame

PR #224, opened 2019-02-02
Implemented a "did you mean" help if a symbol could not be resolved.

hediet:


symtab/src/symtab.rs
```diff
@@ -71,6 +71,10 @@ where
         Ok(())
     }
 
+    pub fn visible_definitions(&self) -> Vec<&S> {
+        self.visible_defs.keys().collect()
```
reiner-dolp: Collecting seems unnecessary to me? You could just return the iterator.
Same goes for the `field_names` getter.

But since we are only calling this in case the compiler aborts because of a detected error, imho this can be merged.

reiner-dolp:
Great work :) 

`bors r=reiner-dolp` as soon as our workflow is fixed.

type_checking/src/method_body_type_checker.rs
```diff
@@ -547,10 +548,49 @@ where
...
+                .chain(self.local_scope.visible_definitions())
+                .chain(self.current_class.field_names())
+                .chain(self.type_system.defined_classes().keys())
+                .chain(self.context.global_vars.keys());
+
+            let mut related: Vec<_> = all_keys
+                .map(|k| (k, k.relation(&var_name.data)))
+                .filter(|(_, r)| r.is_related())
+                .collect();
+            related.sort_by_key(|(_, r)| *r);
```
reiner-dolp: wouldn't sorting by distance make more sense?

symtab/src/symtab.rs
hediet: I thought of both things too. It was a time tradeoff - I would had to figure out how to return the proper iterator trait :D

problame: `... -> impl Iterator<Item=&S>`

reiner-dolp:
force push to rebase onto master, removed merge commit.

bors r+

reiner-dolp:
@hediet we should probably add some tests for this :)

PR #225, opened 2019-02-02
MJLint

flip1995:
First implementation of a linter. (Clippy style :tada:)

I also implemented some example lints to show how easy it is.

A feature to adjust the level for the lints for different code parts would be nice. And now it's here!

```java
//?!allow:lint_name /* global attribute */

//?error:lint_name /* local attribute */
class Foo { }
```

image

The linter is now only invoked when running the compiler with `--check --lint`/`--check -l`.


compiler-cli/src/main.rs
```diff
@@ -366,6 +366,10 @@ macro_rules! until_after_type_check {
             }
         };
 
+        let mut linter = compiler_lib::linter::Linter::default();
```
flip1995: This enables the linter by default when running the compiler with `--check`

flip1995:
@reiner-dolp it seems that the `binary` and `optimization` integration test don't take the reference files into account

reiner-dolp:
When? During compilation? During execution of the reference binary? During execution of the actual binary?

reiner-dolp:
The compiler stdout/stderr default to ignore/empty string. See docs in integration-tests/README. You have to explicitly bind the reference file with "identical-to: file: path"

parser/src/ast.rs
```diff
@@ -235,14 +239,24 @@ impl<'f> ClassMemberKind<'f> {
...
+/// ```
+///
+/// Available attributes are:
+/// - Lint level: `?!level:lint_name` level can be one of `allow`, `info`,
+///   `warning` or `error` e.g `?allow:unused_argument`
+/// - Inline: `?!inline`, `?!noinline`
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum Attribute<'t> {
+    LintLevel(MessageLevel, Spanned<'t, Symbol<'t>>),
+    Inline(bool),
```
flip1995: New Attributes can be added here

parser/src/parser.rs
```diff
@@ -466,11 +504,31 @@ where
                     ast::ClassMemberKind::Field(ty)
                 };
 
-                ast::ClassMember { kind, name }
+                ast::ClassMember { kind, name, attrs }
             })
         })
     }
 
+    fn parse_attribute(&mut self) -> ParserResult<'f, ast::Attribute<'f>> {
```
flip1995: This function needs to be adapted when adding a new attribute.

integration-tests/lints/known_attributes.mj
```diff
@@ -0,0 +1,15 @@
+//?!noinline
```
reiner-dolp: I really don't understand why we have to copy the cryptic syntax of rust. `globally disable/enable LINTNAME` instead of `?!` would be preferable. Same for non-global attributation

integration-tests/lints/unknwon_attribute.mj.stderr
```diff
@@ -0,0 +1,3 @@
+error: expected valid attribute name, found attributes_are_cool at 2:4-22
```
reiner-dolp: why is there no span?

integration-tests/lints/unused_fields.mj.stderr
```diff
@@ -0,0 +1,15 @@
+warning: unused field: `x`. Consider using `_x` instead
```
reiner-dolp: dot at the end of the sentence?

parser/src/parser.rs
```diff
@@ -385,17 +386,53 @@ where
...
+                        {
+                            if let Ok(TokenKind::Operator(Operator::Exclaim)) =
+                                self.lexer.peek_nth(3).map(|t| t.data)
+                            {
+                                self.eat(Exactly(TokenKind::Operator(Operator::Slash)))?;
+                                self.eat(Exactly(TokenKind::Operator(Operator::Slash)))?;
+                                self.eat(Exactly(TokenKind::Operator(Operator::QuestionMark)))?;
+                                self.eat(Exactly(TokenKind::Operator(Operator::Exclaim)))?;
+                                attrs.push(self.parse_attribute()?);
+                                continue;
```
reiner-dolp: NIT: wouldn't this be more readable as a while loop?

reiner-dolp:
IMO there should be a flag that enables or disables the syntax extension, or at least an env var that is set in the `run` script, that is later used by Andreas and Sebastian for benchmarking and standard conformance checks.

parser/src/parser.rs
flip1995: Nope, we need to peek 4 chars ahead before we can decide if it's a global attribute or if it belongs to a class

parser/src/parser.rs
```diff
@@ -453,13 +458,15 @@ where
...
         spanned!(self, {
             let mut attrs = Vec::new();
-            while self
-                .eat_optional(Exactly(TokenKind::Operator(Operator::Slash)))
-                .is_some()
-            {
-                self.eat(Exactly(TokenKind::Operator(Operator::Slash)))?;
-                self.eat(Exactly(TokenKind::Operator(Operator::QuestionMark)))?;
-                attrs.push(self.parse_attribute()?);
+            if std::env::var("VANILLA").is_err() {
```
reiner-dolp: - env var name into a global constant instead of distributing it 3 times over the code base???
- parse_class_member is called several times, caching the lookup?

You can copy paste from here: https://github.com/comprakt/comprakt/commit/7d3708a1d89d3d07618b5cd4c6a096f6e214eb2e

PR #232, opened 2019-02-04
CI: remove --remote from submodule update

reiner-dolp:


reiner-dolp:
Oh, I misread the Travis CI script. This does not make sense as is

PR #234, opened 2019-02-05
Backend Bugs

reiner-dolp:
- also run timeout tests with optimizations
- add tests from feedback


problame:
`integration-tests/timeout/const_optimizable_while.mj` also fails on current master, it fails in `live_variable_analysis.rs`.

@flip1995 any idea what's happening? should we defer the fix in that domain until we have refactored the backend?

problame:
`feedback_5feb_mj` seems to be address, though

PR #235, opened 2019-02-05
only run custom assert logic in integration tests

reiner-dolp:
- only run custom assert logic in integration tests (or if the env var `COMPRAKT_COMPILE_TIME_MJ_ASSERTS` is set)
- abstract custom assert logic introduced by henning, can be applied to any optimization now

Example:
```java
---
optimizations: [ConstantFolding]
expect: change
        
stdout:
    is: "9\n"
---

class ConstantFoldingAfter {
    public void is_const(int expr) {}
}

class ConstantFoldingBefore {
    public void is_not_const(int expr) {}
}

class Opt1After {
    public void is_const(int expr) {}
    public void is_not_const(int expr) {}
}

class CF_Sub {
    ConstantFoldingAfter expect_after = new ConstantFoldingAfter();
    ConstantFoldingBefore expect_before = new ConstantFoldingBefore();

    int a = 14 - 5;

    expect_before.is_not_const(a);
    expect_after.is_const(a);
    /* should not be evaluated since the class name is a mismatch */
    (new Opt1After()).is_const(a);
    (new Opt1After()).is_not_const(a);

    System.out.println(a);
  }
}
```

### DIFF
--- a/optimization/src/compile_time_assertions.rs
+++ b/optimization/src/compile_time_assertions.rs
@@ -9,7 +9,7 @@ use libfirm_rs::{
 use std::fmt;
 
 /// Set this environment variable to enable these special functions
-pub const ENV_VAR_NAME: &'static str = "COMPRAKT_COMPILE_TIME_MJ_ASSERTS";
+pub const ENV_VAR_NAME: &str = "COMPRAKT_COMPILE_TIME_MJ_ASSERTS";
 
 lazy_static::lazy_static! {
     static ref ENABLED: bool = std::env::var(ENV_VAR_NAME).is_ok();
--- a/optimization/src/compile_time_assertions.rs
+++ b/optimization/src/compile_time_assertions.rs
@@ -74,6 +74,31 @@ impl CompileTimeAssertions {
         }
     }
 
+    /// Check if a call is a match for the given test, this allows for
+    /// method overloading by adding a suffix, e.g. a `check_call("is_const",
+    /// assert_const, call, phase)` will match method
+    /// `Assert$M$is_const_int_after`, `Assert$M$is_const_4_before`, ...
+    pub fn check_call<T: Fn(nodes::Call, &Phase)>(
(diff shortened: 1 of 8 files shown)


optimization/src/compile_time_assertions.rs
```diff
@@ -0,0 +1,224 @@
...
+            )
+        );
+
+        let first = nodes[0];
+        for node in &nodes[1..] {
+            assert!(
+                if expect_same {
+                    first == *node
+                } else {
+                    first != *node
```
hediet: Dann wäre `assert.nodeEquality(a, b, c)` gleichbedeutend mit `a != b && a != c`?

optimization/src/compile_time_assertions.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+fn assert_not_const(call: nodes::Call, phase: &Phase) {
+    if let Some(nodes) = check_phase_filter(call, phase) {
+        for node in nodes {
+            assert!(
+                !Node::is_const(node),
+                build_assert_msg(
+                    format!("expected {:?} to NOT be a constant {}", node, phase),
```
hediet: kannst du noch `Spans::span_str(node)` includen?

optimization/src/compile_time_assertions.rs
```diff
@@ -0,0 +1,224 @@
...
+        return false;
+    }
+}
+
+impl CompileTimeAssertions {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn disabled() -> bool {
```
hediet: Ist das teuer aufzurufen?

reiner-dolp: cached now

reiner-dolp:
bors r=hediet

reiner-dolp:
bors r=hediet

PR #237, opened 2019-02-06
integration-tests: add more tests from first feedback round

problame:
Added output to sebastian_20190201__aggressive_3, seems like they
deliberately obfuscated the original test case:
Original array was initialized with length 0, and didn't contain any
output in main, hence the code was not used.


reiner-dolp:
setting fields in the yaml frontmatter to their default value is obviously useless, but it does'nt hurt either.

bors r+

PR #241, opened 2019-02-07
lowering: refactor backend + LIR::Instruction-level live ranges 

problame:
We still need to verify that the live ranges are actually shorter / what the runtime effect on the compiled binaries is. (Expected to relieve register pressure)


problame:
Jenkins tests only fail in binarytree due to memory constraints.

Since all tests pass and it was "just" a refactoring + minor feature, let's get this merged

bors r+

PR #249, opened 2019-02-08
Node Local Optimizations

reiner-dolp:
- [x] substitute power of two divisions by a shift sequence.
- [x] substitute power of two modulo by a shift sequence.
- [x] convert (a + a) to (a << 2)
- [x] substitute power of two multiplications by a shift.


optimization/src/node_local.rs
```diff
@@ -0,0 +1,159 @@
...
+                    log::debug!("Div2Shift: conv {:?}!", conv);
+                    if let Node::Const(divisor) = conv.op() {
+                        log::debug!("Div2Shift: divisor {:?}!", divisor);
+                        if let TarvalKind::Long(divisor_value) = divisor.tarval().kind() {
+                            log::debug!("Div2Shift: divisor value {:?}!", divisor_value);
+                            if is_abs_power_of_two(divisor_value) {
+                                let has_minus = divisor_value < 0;
+
+                                log::debug!("Div2Shift: is power of two!");
+                                let shift_amount = (divisor_value.abs() as f64).log2() as i64;
```
problame: Can this ever not work due to floating point precision? You should be able to just do an abs, then get the highest bit position

problame:
Great work! Wanna do the same for muls with power of two ?

reiner-dolp:
```
class A {
        int i = 0;
        int sum = 0;
        while (i < 2000000001) {
                sum = sum + (i / 2);
                i = i + 1;
        }
        System.out.println(sum);
    }
}
```

image
image

reiner-dolp:
Analog for mod:

image

PR #252, opened 2019-02-09
integration-tests: add crashing bug in constant folding

problame:


problame:
Apparently, this doesn't happen with `--test-threads=1`.

reiner-dolp:
I suspect https://github.com/comprakt/comprakt/pull/249/commits/623e0f206d5e0b644f456792aee02d15c7c34f7f is the fix for this problem? Can someone with the issue try to reproduce it, my system can't.

Issue #253, opened 2019-02-10
Tracking issue: Release 1.0.0


flip1995:
#### can be ignored for 1.0.0
<details>
<summary>PRs</summary>

- #101 (multiple spans) -> unnecessary
- #211 (jenkins) -> unnecessary infra
- #214 (vscode debugging tools) -> not important for release
- #236 (libfirm bug) -> not concerning us
</details>

#### should be looked at
<details>
<summary>PRs</summary>

- #252 (bug in integration tests?) -> fixed in #249
</details>

#### should be merged
- #249 (Div/Mod opt) 

#### can be ignored for 1.0.0
<details>
<summary>issues</summary>

- #23 (read from stdin) -> won't fix
- #51 (TrustedLen Iter) -> too late
- #92 (ord on position) -> we don't support multiple files as input
- #102 (better semantic error) -> unnecessary
- #143 (enum variant) -> bikeshedding
- #151 (sigfpe in runtime) -> unnecessary
- #153 (warn on negative allocations) -> would require CF warnings (or a lint, but this would just address the literal cases)
- #154 (libfirm-rs) -> won't fix
- #155 (integration tests improvement) -> too late
- #162 (debug info) -> won't fix
- #170, #177, #178, #179 (optimizations) -> too late
- #181 (debugger improvement) -> debugger improvement
- #201 (be: lea > SIAM) -> files which would trigger an error here are not possible, since too large
- #212 (improvements of internal timer) -> unnecessary
- #221 (deterministic binaries) -> too late for that big of a refactor
- #244 (Sel/Member nodes) -> won't fix
</details>

#### should be looked at
<details>
<summary>issues</summary>

- #107 (undefined vars, ub) -> already handled
- #109 (ub) -> won't implement
- #216 (optimization order) -> see below
- #242 (inlining heuristics) -> won't implement (inlining is already pretty good without heuristics)
</details>

### Cleanup: :heavy_check_mark: 
<details>
<summary>tasks</summary>

- `assert!` -> `debug_assert!`
- activate following clippy lints in every sub-crate:
  - `print_stdout`
  - `unimplemented`
  - `use_debug`
  - `doc_markdown`
  - `items_after_statements`
  - `match_same_arms`
  - `similar_names`
  - `single_match_else`
  - `use_self`
</details>

### Testing: :heavy_check_mark: 
- Run all integration tests also with moderate 

### Requirements testing: :heavy_check_mark: 
<details>
<summary>tests</summary>

[...]

PR #255, opened 2019-02-10
build: fix run script and make mjtest run like in examination

problame:


flip1995:
cc #253 (requirements testing)

